PR #22, opened 2025-09-05
feat: Add rust-guest check with resolver-std feature in CI workflow

fabriziodemaria:


fabriziodemaria:
Perhaps we don't need to care about this case now since we don't distribute a combined artifact

PR #23, opened 2025-09-05
fix: Fixes in release workflows

fabriziodemaria:
- WASM artifacts were not uploaded/released
- Deployer image was not tagged with the version from Cargo


nicklasl (approved):
LGTM

PR #30, opened 2025-09-16
pass the account id together with the resolver state

vahidlazio:


confidence-resolver/protos/confidence/flags/admin/v1/resolver.proto
```diff
@@ -58,10 +58,13 @@ message ResolverStateUriResponse {
...
 
-// Request to get the resolver state for the whole account
-message ResolverStateRequest {
+message SetResolverStateRequest {
+    bytes state = 1;
+    string account_id = 2;
 }
 
+message ResolverStateRequest {}
+
```
andreas-karlsson, line 67: Why do you want it in this proto?

andreas-karlsson: Ah, maybe you meant to remove this change?

wasm/rust-guest/src/lib.rs
```diff
@@ -1,10 +1,12 @@
 use std::cell::RefCell;
 use std::sync::Arc;
 use std::sync::LazyLock;
-use std::sync::OnceLock;
+
+extern crate alloc;
```
andreas-karlsson, line 5: Is this needed?

andreas-karlsson (approved):

wasm/rust-guest/src/lib.rs
vahidlazio: removed.

PR #32, opened 2025-09-22
feat: js local resolver

andreas-karlsson:


openfeature-provider/js/src/fetch.ts
```diff
@@ -0,0 +1,233 @@
...
+  }
+}
+
+async function bodyRepeater<T extends BodyInit>(body:BodyInit | null | undefined): Promise<() => T> {
+  if(body instanceof ReadableStream) {
+    const blob = await new Response(body).blob();
+    return () => blob.stream() as T
+  }
+  return () => body as T;
+}
```
nicklasl, line 182: AI is suspicious of this function:
>  fetch.ts:176-182 - Type safety issue: bodyRepeater<T extends BodyInit> returns () => T but the generic constraint doesn't guarantee the return type matches, especially for ReadableStream conversion to Blob.

wasm/rust-guest/src/lib.rs
```diff
@@ -6,8 +6,8 @@ use arc_swap::ArcSwapOption;
 use bytes::Bytes;
 use prost::Message;
 
-#[global_allocator]
-static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
+// #[global_allocator]
+// static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
```
nicklasl, line 10: What does this change do?

PR #46, opened 2025-10-09
build: Proper build

andreas-karlsson:
This PR does the following:

1. Creates a better makefile structure with a makefile in each project.
2. Adds a Dockerfile in the root that builds, tests and lints the whole project. This means it's possible to build everything locally without all the different tools installed.
3. Updates CI workflows to use the new docker build with caching. This should make CI more efficient (the docker build does optimal parallelising and caching). We can always run the exact same CI steps locally with just `docker build .`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@
 # ==============================================================================
 # Base image with Rust toolchain (Alpine - more reliable than Debian)
 # ==============================================================================
-FROM rust:1.89-alpine AS rust-base
+FROM alpine:3.22 AS rust-base
 
 # Install system dependencies
 # - protoc/protobuf-dev: Required for prost-build (proto compilation in build.rs)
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,457 @@
...
+COPY --from=rust-deps /usr/local/cargo /usr/local/cargo
+COPY --from=rust-deps /workspace/target /workspace/target
+
+# Copy only Rust-related source files (not Node.js/Python/Java)
+COPY Cargo.toml Cargo.lock ./
+COPY confidence-resolver/ ./confidence-resolver/
+COPY confidence-cloudflare-resolver/ ./confidence-cloudflare-resolver/
+COPY wasm-msg/ ./wasm-msg/
+COPY wasm/rust-guest/ ./wasm/rust-guest/
+COPY wasm/proto/ ./wasm/proto/
```
andreas-karlsson: Use cargo-chef

PR #66, opened 2025-10-17
feat: Java WASM dependency tweaks for dev and release

fabriziodemaria:
🧭 Goal

Ensure the Java provider is correctly released whenever:
	•	Its own implementation changes, or
	•	Its underlying WASM dependencies change.

⚙️ Background

release-please automatically bumps the Java provider version when files change under openfeature-provider/java.
However, the WASM dependency is built and embedded at build time — via the Makefile (locally) or Docker (in CI) — so release-please cannot detect WASM changes by itself.

🧩 Approach

**Local development**
	•	The Makefile always rebuilds the WASM and copies it into the Java resources.
	•	After each build, it updates a marker file in openfeature-provider/java with a unique WASM identifier (e.g. a hash of the built artifact).
	•	When Rust code changes, this causes the marker file to change, creating a diff in openfeature-provider/java.
	•	The developer commits this diff in the same PR as the code change, ensuring release-please picks it up and releases a new Java version.

**CI / Docker build**
	•	CI rebuilds the WASM inside Docker.
	•	Instead of modifying the marker file, CI verifies that the identifier in the checked-in marker file matches the identifier of the freshly built WASM.
	•	This ensures that the embedded WASM committed in the PR is consistent with what CI builds — and prevents missing or outdated WASM artifacts.

✅ Outcome
	•	Any change to the Rust WASM implementation will trigger a version bump in the Java provider.
	•	CI guarantees that the embedded WASM is up-to-date and consistent with the source.
	•	Manual errors (like forgetting to commit the updated WASM) are caught early.
Future PRs will deal with adding the Maven deployments as a next step after the release creations

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -434,6 +434,10 @@ COPY openfeature-provider/java/src ./src/
 # Copy WASM module into resources
 COPY --from=wasm-rust-guest.artifact /confidence_resolver.wasm ../../../wasm/confidence_resolver.wasm
 
+
+# Copy WASM module into resources
+COPY openfeature-provider/java/.embedded-wasm-version ./.embedded-wasm-version
+
 # Set environment variable
 ENV IN_DOCKER_BUILD=1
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -434,6 +434,10 @@ COPY openfeature-provider/java/src ./src/
 # Copy WASM module into resources
 COPY --from=wasm-rust-guest.artifact /confidence_resolver.wasm ../../../wasm/confidence_resolver.wasm
 
+
+# Copy WASM module into resources
+COPY openfeature-provider/java/.embedded-wasm-version ./.embedded-wasm-version
```
fabriziodemaria: We noticed that the HASH of the coped wasm from CI (Docker) is different than the hash we computed locally via Makefile (currently part of this PR's diffs.). We need to investigate

PR #69, opened 2025-10-20
chore(js): formatting with prettier

nicklasl:
split this in two commits to have the formats in a separate one.


openfeature-provider/js/Makefile
```diff
@@ -42,6 +42,7 @@ install: $(INSTALL_STAMP) $(GEN_TS)
 build: $(BUILD_STAMP)
 
 test: $(WASM_ARTIFACT) $(INSTALL_STAMP) $(GEN_TS)
+	yarn format:check
```
fabriziodemaria: I wonder if lint should happen during build rather than test?

PR #104, opened 2025-11-04
fix(java): reload state before creating the initial Resolver

nicklasl:
When adding e2e tests I found a bug where we `sidecarFlagsAdminFetcher.accountId` was null initially which caused a crash when setting this on a proto object.
https://github.com/spotify/confidence-resolver-rust/actions/runs/19079535249/job/54504080290?pr=104#step:5:981

Fixing this by running a `.reload()` before trying to create the initial Resolver.

Also adds tests.


fabriziodemaria (approved):
Nice ⭐️

PR #109, opened 2025-11-05
fix(go): implement StateHandler for proper shutdown

nicklasl:
## Summary
Fixes shutdown flow in the Go OpenFeature provider by implementing the `StateHandler` interface.

## Problem
The OpenFeature Go SDK only calls `Shutdown()` on providers that implement the full `StateHandler` interface (both `Init()` and `Shutdown()` methods). Without this, `openfeature.Shutdown()` would not trigger provider cleanup, leaving logs unflushed and resources unreleased.

## Solution
- Implement `StateHandler` interface with `Init()` and `Shutdown()` methods
- Add `sync.WaitGroup` to prevent race conditions during shutdown
- Ensure ticker goroutine exits immediately on context cancellation
- Proper shutdown sequence: cancel context → wait for goroutines → flush logs → close WASM

## Testing
- Shutdown completes fast with no race conditions or panics

### DIFF
--- a/openfeature-provider/go/confidence/local_resolver_factory.go
+++ b/openfeature-provider/go/confidence/local_resolver_factory.go
@@ -182,6 +182,8 @@ func NewLocalResolverFactory(
 // startScheduledTasks starts the background tasks for state fetching and log polling
 func (f *LocalResolverFactory) startScheduledTasks(parentCtx context.Context) {
 	ctx, cancel := context.WithCancel(parentCtx)
+	f.mu.Lock()
+	defer f.mu.Unlock()
 	f.cancelFunc = cancel
 
 	// Ticker for state fetching and log flushing using StateProvider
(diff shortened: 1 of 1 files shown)


openfeature-provider/go/confidence/local_resolver_factory.go
```diff
@@ -212,15 +216,24 @@ func (f *LocalResolverFactory) startScheduledTasks(parentCtx context.Context) {
...
-	if f.flagLogger != nil {
-		f.flagLogger.Shutdown()
-	}
+	// Close resolver API first (which flushes final logs)
 	if f.resolverAPI != nil {
 		f.resolverAPI.Close(ctx)
+		log.Println("Closed resolver API")
+	}
+	// Wait for background goroutines to exit
+	f.wg.Wait()
```
nicklasl: The wait here ensures that we allow the async timer tasks to complete before we close down the flagLogger.

andreas-karlsson (approved):

PR #123, opened 2025-11-11
feat: add Ruby openfeature provider remote resolver

vahidlazio:


.github/workflows/release-please.yml
```diff
@@ -168,3 +169,36 @@ jobs:
...
+      - name: Setup Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: '3.3'
+
+      - name: Publish to RubyGems
+        env:
+          GEM_HOST_API_KEY: ${{ secrets.RUBYGEMS_API_KEY }}
+        run: |
+          gem push ./artifacts/*.gem
```
nicklasl: Did you consider doing the publish from docker as well? [Like we do in java](https://github.com/spotify/confidence-resolver/pull/123/files#diff-2c84033033d49186c63e6adcd705f63b11ae6814cd76c152c9c486d389fbccf3R131).

JS is a bit special here since OIDC requires that the publish happens from "native"/docker-less npm.

vahidlazio: i try it out 👍

nicklasl (approved):
I haven't checked all the ruby stuff since I don't know it 😸 but the setup LGTM 👍

PR #124, opened 2025-11-11
test: mock-support-server

andreas-karlsson:
A local go server that mocks all required services needed to run a local resolver. Both http and grpc.

### DIFF
--- a/mock-support-server/Dockerfile
+++ b/mock-support-server/Dockerfile
@@ -14,7 +14,7 @@ COPY mock-support-server/genproto ./genproto
 # Build statically linked binary
 RUN --mount=type=cache,target=/go/pkg/mod \
     --mount=type=cache,target=/root/.cache/go-build \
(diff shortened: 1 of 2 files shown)


mock-support-server/main.go
```diff
@@ -0,0 +1,373 @@
...
+	log.Printf("HTTP+h2c (REST+gRPC) listening on %s", httpAddr)
+	srv := &http.Server{Addr: httpAddr, Handler: h2c.NewHandler(handler, &http2.Server{})}
+	if err := srv.ListenAndServe(); err != nil {
+		log.Fatalf("http serve error: %v", err)
+	}
+
+}
+
+// withHTTPLoggingSkipGRPC logs only non-gRPC HTTP requests.
+func withHTTPLoggingSkipGRPC(next http.Handler) http.Handler {
```
andreas-karlsson: It would be better to put this just on the state mux maybe.

PR #147, opened 2025-11-17
feat(openfeature-provider/java): connectionfactory for testing

nicklasl:


openfeature-provider/java/src/main/java/com/spotify/confidence/OpenFeatureLocalResolveProvider.java
```diff
@@ -75,6 +75,53 @@ public OpenFeatureLocalResolveProvider(ApiSecret apiSecret, String clientSecret)
...
+   * LocalProviderConfig config = new LocalProviderConfig(apiSecret, mockFactory);
+   * OpenFeatureLocalResolveProvider provider =
+   *     new OpenFeatureLocalResolveProvider(config, "client-secret");
+   * }</pre>
+   *
+   * @param config the provider configuration including API credentials and optional channel factory
+   * @param clientSecret the client secret for your application, used for flag resolution
+   *     authentication
+   */
+  public OpenFeatureLocalResolveProvider(LocalProviderConfig config, String clientSecret) {
```
fabriziodemaria: Why the client secret is not part of the local config? I suppose the whole credentials system will be changed soon so we don't need to overthink this API now

fabriziodemaria (approved):

PR #150, opened 2025-11-18
feat: Request per second in TelemetryData

fabriziodemaria:
- Added `resolve_rps` to TelemetryData
- WASM-Compatible Timekeeping: use Host::current_time instead of std::time::Instant: the WASM unknown-unknown target (used by Go/Java providers) lacks OS clock access.
- CloudFlare also supports TelemetryData
- Updated Protocol Buffers: Extended the internal API definitions to support resolve_rps, dropped_events, and sdk identity fields.
- GO Provider supports SDK metadata

Tried with the GO demo app (run for 5 seconds, TelemetryData flushed at shutdown):
```
time=2025-11-19T13:15:13.543+01:00 level=INFO msg="Telemetry Data" dropped_events=0 resolve_rps=3052.578186811029 sdk_id=SDK_ID_GO_LOCAL_PROVIDER sdk_version=0.1.0
```

### DIFF
--- a/openfeature-provider/go/confidence/version.go
+++ b/openfeature-provider/go/confidence/version.go
@@ -2,4 +2,4 @@ package confidence
 
 // Version is the current version of the Confidence OpenFeature provider.
 // This version is updated automatically by release-please.
-const Version = "0.1.0"
+const Version = "0.1.0" // x-release-please-version


openfeature-provider/go/confidence/version.go
```diff
@@ -0,0 +1,5 @@
+package confidence
+
+// Version is the current version of the Confidence OpenFeature provider.
+// This version is updated automatically by release-please.
+const Version = "0.1.0"
```
nicklasl: does this need to be marked with `x-release-please-version`?
```suggestion
const Version = "0.1.0" // x-release-please-version
```

PR #161, opened 2025-11-21
feat: send js sdk info in resolve request

nicklasl:
depends on https://github.com/spotify/confidence-resolver/pull/160


openfeature-provider/js/src/version.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ * Version information for the Confidence OpenFeature local provider.
+ * This version is updated automatically by release-please.
+ */
+export const VERSION = '0.1.1'; // x-release-please-version
```
fabriziodemaria: Could it be worth adding a check in CI that this version and the one in `package.json` are aligned?

PR #164, opened 2025-11-24
chore: Streamline proto files setup

fabriziodemaria:
Eliminates proto file duplication by establishing single source of truth in central locations

- Remove duplicate proto files: Delete `openfeature-provider/js/proto/{api,messages}.proto` and `wasm/rust-guest/proto/types.proto`
- Centralize proto definitions: All builds now reference proto files from `wasm/proto/` and `confidence-resolver/protos/` instead of maintaining copies
- Update build systems:
  - JS provider now generates from shared proto files with proper import paths
  - Python host generator enhanced to handle multiple proto paths and create proper package structure
  - Rust guest build simplified to use single proto source
- Consolidate types: Move `ResolveReason` and sticky-related messages to shared `wasm/proto/` location
- Update imports: Adjust TypeScript imports to match new proto structure (e.g., `proto/resolver/api` vs `proto/api`)


andreas-karlsson (commented):
Maybe we should wait with merging this until the big state update PR is through?

PR #170, opened 2025-11-26
refactor!: java materialization interface rework

andreas-karlsson:
## Summary

  Introduces `MaterializationStore` interface to support both sticky assignments and custom targeting
  via materialized segments.

  ## Changes

  - Add `MaterializationStore` interface with `read()` and `write()` operations
  - Define sealed interfaces for read/write operations (`ReadOp`, `WriteOp`, `ReadResult`)
  - Add `MaterializationNotSupportedException` for fallback behavior
  - Update `InMemoryMaterializationRepoExample` to implement new interface (Java 17 compatible)
  - Add comprehensive javadoc to all public APIs
  - Update README with correct terminology and enhanced documentation

  ## Use Cases

  1. **Sticky Assignments**: Maintain consistent variant assignments across evaluations, enable pausing
   intake
  2. **Custom Targeting**: Efficient lookup for materialized segments from precomputed datasets

### DIFF
--- a/openfeature-provider/java/src/main/java/com/spotify/confidence/MaterializationStore.java
+++ b/openfeature-provider/java/src/main/java/com/spotify/confidence/MaterializationStore.java
@@ -60,7 +60,7 @@ public interface MaterializationStore {
    * @throws MaterializationNotSupportedException if the store doesn't support reads (triggers
    *     fallback to remote gRPC resolution)
    */
-  CompletionStage<List<ReadResult>> read(List<? extends ReadOp> ops);
+  CompletionStage<List<ReadResult>> read(List<? extends ReadOp> ops) throws MaterializationNotSupportedException;
 
   /**
    * Performs a batch write of materialization data.
@@ -69,15 +69,15 @@ public interface MaterializationStore {
    * resolution. This includes storing sticky variant assignments and materialized segment
    * memberships. Implementations should be idempotent.
    *
-   * <p><strong>Default Implementation:</strong> Throws {@link UnsupportedOperationException}.
+   * <p><strong>Default Implementation:</strong> Throws {@link MaterializationNotSupportedException}.
    * Override this method if you want to support writing materialization data.
    *
    * @param ops the set of write operations to perform, never null
    * @return a CompletionStage that completes when all writes are finished
-   * @throws UnsupportedOperationException by default if not overridden
+   * @throws MaterializationNotSupportedException by default if not overridden
    */
-  default CompletionStage<Void> write(Set<? extends WriteOp> ops) {
-    throw new UnsupportedOperationException("Unimplemented method 'write'");
+  default CompletionStage<Void> write(Set<? extends WriteOp> ops) throws MaterializationNotSupportedException {
+    throw new MaterializationNotSupportedException();
   }
 
   /**


openfeature-provider/java/src/main/java/com/spotify/confidence/materialization/MaterializationStore.java
```diff
@@ -0,0 +1,14 @@
...
+import java.util.Set;
+import java.util.concurrent.CompletionStage;
+
+public interface MaterializationStore {
+
+  CompletionStage<List<ReadResult>> read(List<? extends ReadOp> ops);
+
+  default CompletionStage<Void> write(Set<? extends WriteOp> ops) {
+    throw new UnsupportedOperationException("Unimplemented method 'write'");
+  }
```
nicklasl: I would rather have this function in a separate interface so we can properly do compile time checks on what we support.

What do you think about having the `read()` in one interface (this one, named `MaterializationStore`) and `write()` in `MaterializationStore.Writer`?

andreas-karlsson: Was thinking about this last night and realized we can use reflection to check `Method.isDefault()`. Then if we want capability knowledge we can do it once at load. With separate interfaces we need to check it on _every use_. But I don't know, maybe there is a better pattern for it. I've lost some enthusiasm for the separate interfaces though. I was also thinking these methods should throw checked exceptions, like `MaterializationRead(Write)NotSupported`

openfeature-provider/java/src/main/java/com/spotify/confidence/SwapWasmResolverApi.java
```diff
@@ -103,94 +99,45 @@ public CompletableFuture<ResolveFlagsResponse> resolveWithSticky(
...
-    missingByUnit.forEach(
-        (unit, materializationInfoItem) -> {
-          materializationInfoItem.forEach(
-              item -> {
-                final Map<String, MaterializationInfo> loadedAssignments;
-                try {
-                  loadedAssignments =
-                      repository
-                          .loadMaterializedAssignmentsForUnit(unit, item.getReadMaterialization())
-                          .get();
```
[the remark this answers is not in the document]
nicklasl: I think the overall changes with how we can handle this flow with this new interface is 👩‍🍳 💋

nicklasl (approved):

openfeature-provider/java/src/main/java/com/spotify/confidence/SwapWasmResolverApi.java
```diff
@@ -76,30 +71,15 @@ public CompletableFuture<ResolveFlagsResponse> resolveWithSticky(
     switch (response.getResolveResultCase()) {
       case SUCCESS -> {
         final var success = response.getSuccess();
-        // Store updates if present
```
fabriziodemaria: Not relevant for this PR, but how do we intend to increase visibility over possible failed store updates?

openfeature-provider/java/src/test/java/com/spotify/confidence/InMemoryMaterializationRepoExample.java
```diff
@@ -1,86 +1,150 @@
...
-
     return CompletableFuture.completedFuture(null);
   }
 
-  @Override
+  /**
+   * Clears all stored materialization data from memory.
+   *
+   * <p>Call this method during application shutdown or test cleanup to free memory.
+   */
```
fabriziodemaria: Should `close` be added to the interface `MaterializationStore`?

openfeature-provider/java/src/main/java/com/spotify/confidence/MaterializationNotSupportedException.java
```diff
@@ -0,0 +1,15 @@
...
+ * <p>This exception triggers the provider to fall back to remote gRPC resolution via the
+ * Confidence service, which manages materializations server-side.
+ *
+ * <p>Users typically don't need to catch this exception - it's handled internally by the
+ * provider's resolution logic.
+ *
+ * @see UnsupportedMaterializationStore
+ * @see MaterializationStore
+ */
+public class MaterializationNotSupportedException extends RuntimeException {}
```
fabriziodemaria: I wonder if this should be a checked exception? It would guarantee that we protect against this (expected) scenario

openfeature-provider/java/src/test/java/com/spotify/confidence/ResolveTest.java
```diff
@@ -292,8 +360,59 @@ private static byte[] buildResolverStateBytes(Map<String, Flag> flagsMap) {
...
-            .setClientSecret(TestBase.secret)
+            .setClientSecret(secret)
             .build());
     return builder.build().toByteArray();
   }
+
+  private ResolveFlagsResponse resolveWithContext(
+      List<String> flags, String username, Struct struct, boolean apply, String secret) {
+    return resolverApi
+        .resolveWithSticky(
```
fabriziodemaria: nit: I suggest we have a single `resolve` signature with optional stickiness in the request. Not for this PR

openfeature-provider/java/src/main/java/com/spotify/confidence/OpenFeatureLocalResolveProvider.java
```diff
@@ -321,14 +312,19 @@ public ProviderEvaluation<Value> getObjectEvaluation(
...
-                      .setFailFastOnSticky(getFailFast(stickyResolveStrategy))
-                      .build())
-              .get();
+      try {
+        resolveFlagResponse =
+            wasmResolveApi
+                .resolveWithSticky(
+                    ResolveWithStickyRequest.newBuilder()
+                        .setResolveRequest(req)
+                        .setFailFastOnSticky(false)
```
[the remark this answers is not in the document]
nicklasl: I don't think so but I'm hesitant to change the proto. Especially since they are shared, or?

openfeature-provider/java/src/main/java/com/spotify/confidence/SwapWasmResolverApi.java
```diff
@@ -76,30 +71,15 @@ public CompletableFuture<ResolveFlagsResponse> resolveWithSticky(
...
-          final var currentRequest =
-              handleMissingMaterializations(
-                  request, missingMaterializations.getItemsList(), repository);
-          return resolveWithSticky(currentRequest);
-        }
-
-        throw new RuntimeException(
-            "Unknown sticky resolve strategy: " + stickyResolveStrategy.getClass());
+        return handleMissingMaterializations(request, missingMaterializations)
+            .thenCompose(this::resolveWithSticky);
```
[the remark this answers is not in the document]
fabriziodemaria: Yes, we don't expect e third re-entry, so if that happens it's a bug and I'd say we return a general error

openfeature-provider/java/src/main/java/com/spotify/confidence/OpenFeatureLocalResolveProvider.java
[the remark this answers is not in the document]
fabriziodemaria: I am not worry about breaking the API at this time. I am a bit worried that these options are a bit confusing at first sight, and if we are not using them now it's probably good to strip them out. We don't need to do it in this PR

openfeature-provider/java/src/main/java/com/spotify/confidence/RemoteResolver.java
```diff
@@ -4,6 +4,8 @@
 import com.spotify.confidence.flags.resolver.v1.ResolveFlagsResponse;
 import java.util.concurrent.CompletableFuture;
 
-public non-sealed interface ResolverFallback extends StickyResolveStrategy {
+public interface RemoteResolver {
```
andreas-karlsson: Do we need this interface? I don't mind if we keep it, but maybe make it package private?

PR #193, opened 2025-12-03
refactor: Compositional Local Resolver

andreas-karlsson:
## Compositional Local Resolver Refactoring

  Restructures the Go provider with improved architecture and organization:

  - **Compositional design**: Refactored local resolver to use compositional patterns for better modularity
  - **Internal packages**: Organized code into `internal/flag_logger` and `internal/local_resolver` packages
  - **Simplified async logging**: Removed chunking complexity, fire-and-forget writes without error returns
  - **WASM relocation**: Moved assets to `internal/local_resolver/assets/`
  - **Dockerfile sync**: Updated build to reflect new structure

### DIFF
--- a/openfeature-provider/go/confidence/provider_test.go
+++ b/openfeature-provider/go/confidence/provider_test.go
@@ -489,7 +489,7 @@ func TestLocalResolverProvider_Init_NilStateProvider(t *testing.T) {
 	provider := NewLocalResolverProvider(
 		mockResolverSupplier,
 		nil, // nil state provider
-		nil,
+		&tu.MockFlagLogger{},
 		"secret",
 		nil,
 	)
@@ -508,7 +508,7 @@ func TestLocalResolverProvider_Init_NilResolverAPI(t *testing.T) {
 	provider := NewLocalResolverProvider(
 		nil, // nil resolver API
 		&tu.StateProviderMock{},
-		nil,
+		&tu.MockFlagLogger{},
 		"secret",
 		nil,
 	)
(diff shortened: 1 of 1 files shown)


openfeature-provider/go/confidence/internal/local_resolver/pool.go
```diff
@@ -0,0 +1,119 @@
...
+	})
+}
+
+func (s *PooledResolver) maintenance(fn func(LocalResolver) error) error {
+	errs := []error{}
+	s.mmu.Lock()
+	defer s.mmu.Unlock()
+	for i, slot := range s.slots {
+		func() {
+			slot.rw.Lock()
```
nicklasl: I talked to AI on the topic of congestion between the maintenance locks and the resolve locks. And wether it makes sense to reverse one of the for loops to reduce the risk for contention but "we" think the benefit would be marginal.

nicklasl (approved):
Looks good. Great improvement!

But how is this breaking? Because we are moving types to internal? -- Since none of the types were meant to be public to begin with I would argue we let that one slip through (meaning we don't mark this as breaking).

openfeature-provider/go/confidence/internal/local_resolver/pool.go
andreas-karlsson: Yes. I had one of the iteration orders reversed before, but I didn't see any difference, so went with simpler code. And it makes sense. The write lock is held a lot(!) longer, and readers will zip past it with a _try-read-lock_ so whether they iterate in the same direction or not doesn't matter.

PR #201, opened 2025-12-09
feat: add support for materialized segment targeting criteria

nicklasl:
## Summary

  Add support for materialized segment targeting criteria with lazy evaluation during flag resolution.

  ## Changes

  ### Core Implementation (Rust)
  - Add `MaterializedSegmentCriterion` to targeting criteria types
  - Implement lazy evaluation: only request materializations when a rule requiring them is encountered during evaluation
  - If an earlier rule matches without materializations, resolution succeeds without checking later rules
  - Error propagation: missing materialization data triggers `MissingMaterializations` error
  - Add helper methods: `collect_missing_materializations_for_flag()`, `find_materialized_segments_in_segment()`
  - Support inclusion-based segment matching via `InclusionReadOp`

  ### Protocol Buffers
  - Add `target.proto` with `MaterializedSegmentCriterion` definition
  - Update `resolver.proto` for materialization operations

  ### Go Provider
  - Add comprehensive integration tests for materialized segment resolution
  - Test coverage for both inclusion and exclusion cases


fabriziodemaria (approved):
I find the evaluation logic a bit difficult to follow now. It seems to me that the "metarialization" is a bit overloaded now: we have `Rule.materialization_spec`, which is used for sticky variants, but also `Segment.targeting.criteria / MaterializedSegmentCriterion` which is used for inclusion-check only. 
Perhaps the naming will make more sense to me once I get used to it.

Overall the PR looks functional to me and there is no blocking comment on my side, approved

PR #204, opened 2025-12-10
refactor: change how and from where Go generates protos

nicklasl:


openfeature-provider/go/scripts/generate_proto.sh
```diff
@@ -23,182 +23,34 @@ if ! command -v protoc-gen-go-grpc &> /dev/null; then
...
 fi
 
-# Generate protobuf Go files
 echo "Generating protobuf Go files..."
 
 mkdir -p confidence/proto
 
 # Generate wasm messages proto (WASM-specific types)
+echo "Generating WASM messages..."
 protoc --proto_path=../../wasm/proto \
```
andreas-karlsson: We could copy what's needed here into `open_feature/proto` we should then use the same file file path(s) as in wasm/proto

PR #205, opened 2025-12-11
refactor: change internal API for missing materializations

nicklasl:


nicklasl:
I'm splitting this into multiple PR's to extract some of the "chores"

PR #211, opened 2025-12-15
feat: pluggable materialization

andreas-karlsson:
Introduces a MaterializationStore interface and makes the Confidence remote materialization store available.

### DIFF
--- a/openfeature-provider/js/src/ConfidenceServerProviderLocal.ts
+++ b/openfeature-provider/js/src/ConfidenceServerProviderLocal.ts
@@ -77,13 +77,19 @@ export class ConfidenceServerProviderLocal implements Provider {
           ],
           'https://resolver.confidence.dev/*': [
             withRouter({
-              '*/v1/materialization:readMaterializedOperations': [withTimeout(200 * TimeUnit.MILLISECOND)],
+              '*/v1/materialization:readMaterializedOperations': [
+                withRetry({
+                  maxAttempts: 3,
+                  baseInterval: 100,
+                }),
+                withTimeout(0.5 * TimeUnit.SECOND),
+              ],
               '*/v1/materialization:writeMaterializedOperations': [
                 withRetry({
                   maxAttempts: 3,
                   baseInterval: 100,
                 }),
-                withTimeout(1 * TimeUnit.SECOND),
+                withTimeout(0.5 * TimeUnit.SECOND),
               ],
               '*/v1/clientFlagLogs:write': [
                 withRetry({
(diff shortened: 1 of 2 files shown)


openfeature-provider/js/src/ConfidenceServerProviderLocal.ts
```diff
@@ -62,7 +70,7 @@ export class ConfidenceServerProviderLocal implements Provider {
           ],
           'https://resolver.confidence.dev/*': [
             withRouter({
-              '*/v1/flags:resolve': [
+              '*/v1/materialization:readMaterializedOperations|*/v1/materialization:writeMaterializedOperations': [
```
nicklasl: I think it is reasonable to split these and allow for configuration of timeout individually.
in go and java we allow for env var configuration of this (defaulting to deadline 2 sec for read and 5 sec for write).

openfeature-provider/js/src/ConfidenceServerProviderLocal.ts
```diff
@@ -119,75 +127,60 @@ export class ConfidenceServerProviderLocal implements Provider {
...
-        apply: true,
-        clientSecret: this.options.flagClientSecret,
-        sdk: {
-          id: SdkId.SDK_ID_JS_LOCAL_SERVER_PROVIDER,
-          version: VERSION,
+    try {
+      const [flagName, ...path] = flagKey.split('.');
+
+      // Build resolve request
+      // Always use sticky resolve request with remote fallback
```
nicklasl:
```suggestion
      // Always use sticky resolve request
```

openfeature-provider/js/src/ConfidenceServerProviderLocal.ts
```diff
@@ -119,75 +127,60 @@ export class ConfidenceServerProviderLocal implements Provider {
...
+      const { results: materializations } = await this.readMaterializations(stickyResponse.readOpsRequest);
+      stickyResponse = this.resolver.resolveWithSticky({ ...stickyRequest, materializations });
     }
 
-    // Handle missing materializations by falling back to remote resolver
-    if (response.readOpsRequest) {
-      return await this.remoteResolve(request.resolveRequest!);
+    if (!stickyResponse.success) {
+      // this shouldn't happen with failFast = false. Although it _could_ happen if the state changed and added a new read
+      throw new Error('Missing materializations');
```
nicklasl: makes sense to handle this here.

openfeature-provider/js/src/ConfidenceServerProviderLocal.ts
andreas-karlsson:
```suggestion
```

openfeature-provider/js/src/ConfidenceServerProviderLocal.ts
andreas-karlsson: I'm happy to split them. But I'd prefer options rather than env vars? Do we want that? And/or do we also want separate defaults?

PR #213, opened 2025-12-19
build(js): detect bundle split

nicklasl:


andreas-karlsson (approved):
I'd prefer if we also did this check locally, but I guess that can be added later.

PR #223, opened 2026-01-15
feat: add rust provider

vahidlazio:


openfeature-provider/rust/src/provider.rs
```diff
@@ -0,0 +1,725 @@
...
+    /// Use a custom materialization store.
+    Custom(Arc<dyn MaterializationStore>),
+}
+
+/// Configuration options for the Confidence provider.
+pub struct ProviderOptions {
+    /// The client secret for authentication.
+    pub client_secret: String,
+    /// Timeout for initialization in milliseconds.
+    pub initialize_timeout_ms: Option<u64>,
```
andreas-karlsson: Doesn't rust have some `Duration` type?

PR #232, opened 2026-01-20
fix: align rust provider component name in release-please

nicklasl:
## Summary
- Add `component` field to rust provider in release-please config to match naming convention used by other providers

This ensures the rust provider shows as `openfeature-provider/rust` in release notes instead of the crate name `spotify-confidence-openfeature-provider-local`.


nicklasl:
I hope this'll do the trick, I didn't try it locally.

PR #235, opened 2026-01-20
feat(wasm): add wasm API to apply previously resolved flags

nicklasl:
This PR adds a new API in wasm to apply flags that were previously resolved.


confidence-resolver/src/lib.rs
```diff
@@ -2086,6 +2086,241 @@ mod tests {
...
+                version: "0.1.0".to_string(),
+            }),
+        };
+
+        // Verify that NO assignment was logged before apply_flags
+        assert_eq!(
+            TestLogger::get_logs().len(),
+            0,
+            "No assignments should be logged before apply_flags"
+        );
```
andreas-karlsson: This one or the similar check on L2188 is unnecessary. We haven't had any interaction in between?

PR #251, opened 2026-01-27
ci(js): add typecheck to make test and fix type errors

nicklasl:
## Summary
- Add typecheck step to `make test` in the JS provider
- Fix type errors in flag logging e2e tests


openfeature-provider/js/Makefile
```diff
@@ -43,6 +43,7 @@ build: $(BUILD_STAMP)
 
 test: $(WASM_ARTIFACT) $(INSTALL_STAMP) $(GEN_TS)
 	yarn format:check
+	yarn typecheck
```
andreas-karlsson: Should this be part of build instead?

PR #283, opened 2026-02-18
feat: Support for startsWith and endsWith

fabriziodemaria:
Additionally:
- Returns evaluation error is rule not recognized
- Updates the Reason string in the Java provider to match the OpenFeature enum

### DIFF
--- a/openfeature-provider/java/src/main/java/com/spotify/confidence/sdk/OpenFeatureLocalResolveProvider.java
+++ b/openfeature-provider/java/src/main/java/com/spotify/confidence/sdk/OpenFeatureLocalResolveProvider.java
@@ -5,7 +5,6 @@ import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.google.protobuf.Struct;
 import com.spotify.confidence.sdk.flags.resolver.v1.ResolveFlagsRequest;
 import com.spotify.confidence.sdk.flags.resolver.v1.ResolveFlagsResponse;
-import com.spotify.confidence.sdk.flags.resolver.v1.ResolveReason;
 import com.spotify.confidence.sdk.flags.resolver.v1.ResolveWithStickyRequest;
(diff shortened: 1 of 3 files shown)


openfeature-provider/java/src/main/java/com/spotify/confidence/sdk/OpenFeatureLocalResolveProvider.java
```diff
@@ -449,6 +450,23 @@ public ProviderEvaluation<Value> getObjectEvaluation(
...
+        return Reason.DEFAULT.toString();
+      case RESOLVE_REASON_FLAG_ARCHIVED:
+        return Reason.DISABLED.toString();
+      case RESOLVE_REASON_TARGETING_KEY_ERROR:
+      case RESOLVE_REASON_ERROR:
+      case RESOLVE_REASON_UNRECOGNIZED_TARGETING_RULE:
+        return Reason.ERROR.toString();
+      default:
+        return Reason.UNKNOWN.toString();
+    }
```
fabriziodemaria: Reason from the Provider matches the values expected by OpenFeature. Kinda unrelated to the PR scope, but good to have the correct "ERROR" reason string when `RESOLVE_REASON_UNRECOGNIZED_TARGETING_RULE` happens

fabriziodemaria: My understanding of the OF spec is that ResolveReason can be REASON_ERROR, then the ERROR_CODE + ERROR_MESSAGE define the specific error (where REASON and ERROR_CODE are OF-specific texts, not sure why they don't enforce them as enums).

However I notice that applying this across all the provider becomes a bigger task. I added a commit that reverts this change and simply returns RESOLVE_REASON_UNRECOGNIZED_TARGETING_RULE as Reason (same as, for example, RESOLVE_REASON_TARGETING_KEY_ERROR). We can fix the error mapping in a future PR

PR #345, opened 2026-03-19
fix(resolver): bucket-match materialized assignments with duplicate variants

nicklasl:
## Summary
- Adds a spec test case where two assignments in a materialization rule point to the same variant but have different bucket ranges and assignment IDs
- Fixes `select_assignment` to use bucket-based matching (same as the Java resolver) when multiple assignments share a variant, instead of always returning the first match

## Test plan
- [x] `cargo test -p confidence_resolver mat_same_variant` passes
- [x] All 370 unit tests pass
- [x] WASM synced to Go provider


yyoyoian-pixel (approved):
LGTM — targets a real and subtle edge case (two assignments for the same variant in different bucket ranges). The new `assignment_id` assertion infrastructure via resolve token decryption is a useful addition to the spec test framework.

Optional suggestion: a complementary test with a different `targeting_key` that hashes into the 0-500 range (expecting `a1`) would fully exercise both assignments and guard against boundary bugs.

vahidlazio (approved):
LGTM — targets a real and subtle edge case (two assignments for the same variant in different bucket ranges). The new `assignment_id` assertion infrastructure via resolve token decryption is a useful addition to the spec test framework.

Optional suggestion: a complementary test with a different `targeting_key` that hashes into the 0-500 range (expecting `a1`) would fully exercise both assignments and guard against boundary bugs.

PR #369, opened 2026-04-14
fix: free WASM guest request allocation in host functions

nicklasl:
## Summary

- Fix memory leak where `wasm_msg_host_current_time` host function never frees the request pointer allocated by the WASM guest
- Each `resolve_flags`/`apply_flags` call leaked ~20 bytes of WASM linear memory (which can only grow, never shrink)
- Over days of continuous operation this caused unbounded memory growth (observed 50MB → 300MB over a week)
- Go, JS, and Python host functions now free the incoming request pointer. Java was already correct.
- Add WASM memory regression tests for all four providers

## Test plan

- [x] Go: `TestWasmMemoryStableOnRepeatedResolveCalls` — 50k resolves, zero memory growth
- [x] JS: `wasm memory stability` — 50k resolves, zero memory growth
- [x] Python: `test_multiple_resolves_dont_leak_memory` — 50k resolves, zero memory growth
- [x] Java: `wasmMemoryStableOnRepeatedResolveCalls` — 50k resolves, zero memory growth

### DIFF
--- a/openfeature-provider/go/confidence/internal/local_resolver/wasm.go
+++ b/openfeature-provider/go/confidence/internal/local_resolver/wasm.go
@@ -2,15 +2,13 @@ package local_resolver
 
 import (
 	"context"
+	_ "embed"
 	"encoding/binary"
 	"errors"
 	"fmt"
 	"log/slog"
 	"sync"
 	"sync/atomic"
-	"time"
-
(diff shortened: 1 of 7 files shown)


andreas-karlsson (approved):
Nice!

PR #383, opened 2026-04-24
fix(resolver): skip targeting_key validation when rule has blank selector

nicklasl:
## Summary
- When a rule's `targetingKeySelector` is empty, an invalid `targeting_key` value (e.g. a struct) is now treated as absent rather than raising `TARGETING_KEY_ERROR`. Full-rollout rules match; partial-rollout rules skip. Aligns with Java resolver behaviour.
- Removes the `"rust"` override from the `full_rollout_invalid_key_type` spec test — the `general` expectation (`MATCH`) now applies.
- WASM re-synced into the Go provider.

Closes #339

## Test plan
- [x] `make -C confidence-resolver test` (343/343 green)
- [x] `make -C confidence-resolver lint` (clippy clean)
- [x] `make sync-wasm-go`


yyoyoian-pixel (approved):
LGTM — small, well-motivated fix that correctly aligns the Rust resolver with the Java resolver for blank selectors on full rollouts.

Two minor suggestions (non-blocking):

1. **Consider adding a spec test for the partial-rollout case** — blank selector + partial rollout + invalid key type should still return `TARGETING_KEY_ERROR`. This would lock in the intentional asymmetry and prevent future regressions.

2. **Optional comment in the `Err(_)` arm** — the behavior differs from the `Ok(None)` arm for partial rollouts (missing key silently skips, but wrong type errors). A brief note like:
   ```rust
   // Note: unlike Ok(None) which skips silently for partial rollouts,
   // Err (wrong type) still errors — a malformed key signals a caller bug.
   ```
   would help future readers understand the intentional difference.

vahidlazio (approved):
LGTM — small, well-motivated fix that correctly aligns the Rust resolver with the Java resolver for blank selectors on full rollouts.

Two minor suggestions (non-blocking):

1. **Consider adding a spec test for the partial-rollout case** — blank selector + partial rollout + invalid key type should still return `TARGETING_KEY_ERROR`. This would lock in the intentional asymmetry and prevent future regressions.

2. **Optional comment in the `Err(_)` arm** — the behavior differs from the `Ok(None)` arm for partial rollouts (missing key silently skips, but wrong type errors). A brief note like:
   ```rust
   // Note: unlike Ok(None) which skips silently for partial rollouts,
   // Err (wrong type) still errors — a malformed key signals a caller bug.
   ```
   would help future readers understand the intentional difference.

PR #386, opened 2026-04-27
fix: surface apply_flags errors instead of swallowing

nicklasl:
## Summary

- WASM guest's `apply_flags` previously discarded all errors via `let _ = resolver.apply_flags(&request)`, so a malformed `ApplyFlagsRequest` (e.g. missing top-level `send_time`, or a `client_secret` that has been rotated out of state) silently produced zero `FlagAssigned` entries on the next flush — appearing to consumers as a successful apply that never reached the backend.
- The guest now propagates the underlying error. Each provider (Java, Go, JS) catches it, logs a WARN, and continues. Apply remains best-effort logging — no caller is crashed — but failures are now diagnosable.

## Changes

| Component | Change |
|---|---|
| `wasm/rust-guest/src/lib.rs` | `apply_flags` returns `Err` instead of swallowing |
| `openfeature-provider/java/.../WasmLocalResolver.java` | Catch + `logger.warn(...)` around `consumeResponse` |
| `openfeature-provider/go/.../local_resolver/wasm.go` | Catch + `slog.Warn(...)`, returns nil (mirrors `RegisterResolve`) |
| `openfeature-provider/js/src/WasmResolver.ts` | Catch non-`RuntimeError` + `logger.warn(...)`, no rethrow |
| `openfeature-provider/go/.../assets/confidence_resolver.wasm` | Rebuilt WASM (Go embeds it) |
| Java provider's `src/main/resources/wasm/...` | Not committed (gitignored, built locally) |

Python provider is unaffected — it has no separate apply path; resolves run with `apply=true` inline.

## Test plan

- [x] New unit test `ApplyFlagsErrorHandlingTest` covering three scenarios:
  - Malformed request (no `send_time`) → swallowed, 0 `FlagAssigned`, WARN log with `"send_time is required"`
  - Well-formed request → 1 `FlagAssigned` with correct `resolveId`, `clientInfo`, `targetingKey`, `assignmentInfo` (segment, variant), `assignmentId`, `applyTime`; no WARN
  - State rotated between resolve and apply (secret no longer present) → swallowed, 0 `FlagAssigned`, WARN log with `"client secret not found"`
- [x] Full Java test suite: 89/89 pass
- [x] Go unit tests: pass
- [ ] JS tests: not yet exercised locally


vahidlazio (commented):
Nice change — surfacing these errors is a clear improvement across all providers. One issue in the JS provider:

## JS `RuntimeError` branch is now silently swallowed

The old code rethrew all errors after handling `RuntimeError`, so callers saw the error. The new code logs non-WASM errors (great!) but the `RuntimeError` branch now does neither — no log, no rethrow. A WASM crash during `apply_flags` is now *less* visible than before.

**Suggested fix** — add `logger.warn` in the `RuntimeError` branch too:

```typescript
} catch (error: unknown) {
    if (error instanceof WebAssembly.RuntimeError) {
        logger.warn('Failed to apply flags (WASM RuntimeError, reloading instance):', error);
        this.reloadInstance(error);
    } else {
        logger.warn('Failed to apply flags:', error);
    }
}
```

Keeps the best-effort contract (no rethrow) while ensuring WASM crashes are visible in logs, consistent with Go and Java.

openfeature-provider/java/src/main/java/com/spotify/confidence/sdk/WasmLocalResolver.java
```diff
@@ -171,7 +171,12 @@ public void applyFlags(ApplyFlagsRequest request) {
       }
       final int reqPtr = transferRequest(request);
       final int respPtr = (int) wasmMsgGuestApplyFlags.apply(reqPtr)[0];
-      consumeResponse(respPtr, ApplyFlagsResponse::parseFrom);
+      try {
+        consumeResponse(respPtr, ApplyFlagsResponse::parseFrom);
+      } catch (RuntimeException e) {
+        // Apply is best-effort logging — surface the failure but don't propagate.
+        logger.warn("Failed to apply flags: {}", e.getMessage());
+      }
```
fabriziodemaria, line 179: The inner try/catch only wraps `consumeResponse`, so the `wasmMsgGuestApplyFlags.apply(reqPtr)` call above it can still throw (Chicory `WasmRuntimeException` on trap, OOM in `wasm_msg_alloc`, etc.) and propagate out of `applyFlags` despite the "don't propagate" comment.

Compare:
- JS catches `WebAssembly.RuntimeError` separately and reloads the instance, then swallows.
- Go is wrapped by `RecoveringResolver`, which catches the panic from `r.call`'s `panic(err)`.
- Java has neither an outer catch nor a recovering wrapper here.

If "apply is best-effort logging" is the contract, consider widening the catch to include the WASM call (or adding a recovery wrapper analogous to Go) so traps are also swallowed.

nicklasl: Thanks for the close read! One correction: Java does have a recovering wrapper — `RecoveringResolver` sits between `PooledResolver` and `WasmLocalResolver`, catches `ChicoryException`, kicks off a background recreate, then rethrows (analogous to Go's `withRecover`).

The composition is:
```
FlagResolverService → OpenFeatureLocalResolveProvider → MaterializingResolver
  → PooledResolver → RecoveringResolver → WasmLocalResolver
```

So the narrow scope of the inner try here is deliberate: it only handles the `Response.error` path this PR introduces (caller mistakes like missing `send_time`, rotated `client_secret`). Traps from `wasmMsgGuestApplyFlags.apply` are meant to bubble up so `RecoveringResolver` can see them and trigger recreate — widening the catch here would blind the recovery layer to traps and the instance would stay broken.

nicklasl:
@vahidlazio good eye — but `reloadInstance` already does `logger.error('Failure calling into wasm:', error)` at the top of its body (line 194), so the WASM crash isn't actually silent, just logged with a generic message rather than an apply-specific one. The `reloadInstance` path also reloads the instance, similar to Java/Go's `RecoveringResolver` flow.

fabriziodemaria (approved):
What happens with `Rust` and `Python` if they don't swallow the new exception (I don't see them in this pr)? It would be great if WASM/all-languages had checked exception to ensure this is handled, but I guess this is not possible.

Approving!

nicklasl:
> What happens with `Rust` and `Python` if they don't swallow the new exception (I don't see them in this pr)? It would be great if WASM/all-languages had checked exception to ensure this is handled, but I guess this is not possible.
> 
> Approving!

I believe we have not yet added the apply function on those providers.

PR #393, opened 2026-05-05
feat(js): Next.js Pages Router support

andreas-karlsson:
## Summary

Adds three subexports under `./pages-router/*` for the Next.js Pages Router, parallel to the existing `react-server`/`react-client` pair for App Router:

- **`pages-router/server`** — `withConfidence(config, gssp?)` HOC that resolves the flag bundle in `getServerSideProps` (no exposure) and merges it into `pageProps`. Plus a low-level `resolveConfidence` helper.
- **`pages-router/client`** — `<ConfidencePagesProvider>` for `_app.tsx`. The existing `useFlag` / `useFlagDetails` from `react-client` work unchanged.
- **`pages-router/api`** — `applyHandler({ providerName? }?)` factory for the `/api/confidence/apply` POST route the client uses to log exposure.

The resolve token is sealed with AES-256-GCM (key from `CONFIDENCE_TOKEN_KEY` env var) before reaching the client, since Pages Router has no encrypted-closure equivalent to App Router server actions. The apply route opens it server-side.

`README-REACT.md` gets a new "Next.js Pages Router" section. The App Router quick-start was also moved to `instrumentation.ts` since that's router-agnostic and avoids the build-time side-effect issue with top-level-await setup files.

## Test plan

- [ ] `yarn typecheck` clean
- [ ] `yarn test` — 7 token tests + 6 apply-handler tests pass
- [ ] `yarn build` emits `dist/pages-router/{server,client,api}.{js,d.ts}`
- [ ] Existing `react-server` / `react-client` builds unchanged
- [ ] Verified end-to-end against a Pages Router demo app (server-side resolve → client hook reads → apply route fires exposure)

### DIFF
--- a/openfeature-provider/js/src/pages-router/api.ts
+++ b/openfeature-provider/js/src/pages-router/api.ts
@@ -45,6 +45,13 @@ export function applyHandler(opts: ApplyHandlerOptions = {}): NextApiHandler {
       return;
     }
 
+    // Error bundles carry an empty resolveToken (see FlagBundle.error); skip
+    // applyFlag in that case to match the App Router's `!bundle.errorCode` gate.
+    if (!openedToken) {
+      res.status(204).end();
+      return;
+    }
+
     provider.applyFlag(openedToken, body.flagName);
     res.status(204).end();
   };
--- a/openfeature-provider/js/src/pages-router/api.ts
+++ b/openfeature-provider/js/src/pages-router/api.ts
@@ -1,0 +1,51 @@
+import type { NextApiHandler } from 'next';
+import { getConfidenceProvider } from './common';
+import { openResolveToken } from './token';
+
+export interface ApplyHandlerOptions {
+  /** Use a non-default OpenFeature provider by name. */
+  providerName?: string;
+}
+
+/**
+ * Builds the POST handler that the client `ConfidencePagesProvider` calls to
+ * fire exposure events. Mount it at `/api/confidence/apply` (or pass a custom
+ * `apiPath` to `ConfidencePagesProvider` if you mount it elsewhere).
+ *
+ * @example
+ * // pages/api/confidence/apply.ts
+ * import { applyHandler } from '@spotify-confidence/openfeature-server-provider-local/pages-router/api';
+ * export default applyHandler();
+ */
+export function applyHandler(opts: ApplyHandlerOptions = {}): NextApiHandler {
+  return async (req, res) => {
+    if (req.method !== 'POST') {
+      res.setHeader('Allow', 'POST');
+      res.status(405).end();
+      return;
+    }
+
+    const body = req.body as { resolveToken?: unknown; flagName?: unknown } | undefined;
+    if (!body || typeof body.resolveToken !== 'string' || typeof body.flagName !== 'string') {
+      res.status(400).end();
+      return;
+    }
+
+    const provider = getConfidenceProvider(opts.providerName);
+    if (!provider) {
+      res.status(503).end();
+      return;
+    }
+
+    let openedToken: string;
+    try {
+      openedToken = openResolveToken(body.resolveToken);
+    } catch {
+      res.status(400).end();
+      return;
+    }
+
+    provider.applyFlag(openedToken, body.flagName);
+    res.status(204).end();
+  };
+}
--- a/openfeature-provider/js/src/pages-router/client.tsx
+++ b/openfeature-provider/js/src/pages-router/client.tsx
@@ -1,0 +1,59 @@
+'use client';
+
+import { useCallback, type ReactNode } from 'react';
+// Package self-reference so the pages-router/client bundle does NOT inline
+// react/client — otherwise we'd ship two copies of `ConfidenceContext` and
+// `<ConfidencePagesProvider>` would write to a different context than the one
+// `useFlag` / `useFlagDetails` read from. The consumer's bundler resolves
+// this via the package's `./react-client` exports entry and dedupes it against
+// any other usage of react-client in the same app.
+import { ConfidenceClientProvider } from '@spotify-confidence/openfeature-server-provider-local/react-client';
+import { DEFAULT_APPLY_PATH } from './constants';
+import type { ConfidencePageProps } from './types';
+
+export type { ConfidencePageProps } from './types';
+
+interface Props {
+  /**
+   * The Confidence payload returned from `withConfidence` in
+   * `getServerSideProps`, normally pulled out of `pageProps` in `_app.tsx`.
+   * Pages that don't resolve flags pass `undefined` and any `useFlag` /
+   * `useFlagDetails` calls in their tree return defaults.
+   */
+  confidence?: ConfidencePageProps;
+  /** Override the apply API route. Must match where you mount `applyHandler`. */
+  apiPath?: string;
+  children: ReactNode;
+}
+
+/**
+ * Place at the top of `_app.tsx`. Bridges the bundle resolved on the server
+ * to the client `useFlag` / `useFlagDetails` hooks.
+ */
+export function ConfidencePagesProvider({
+  confidence,
+  apiPath = DEFAULT_APPLY_PATH,
+  children,
+}: Props): React.ReactElement {
+  const resolveToken = confidence?.resolveToken;
+
+  const apply = useCallback(
+    async (flagName: string) => {
+      if (!resolveToken) return;
+      await fetch(apiPath, {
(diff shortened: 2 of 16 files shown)


openfeature-provider/js/src/pages-router/server.ts
```diff
@@ -0,0 +1,69 @@
...
+  config: WithConfidenceConfig,
+  ctx: GetServerSidePropsContext,
+): Promise<ConfidencePageProps> {
+  const provider = getConfidenceProvider(config.providerName);
+  if (!provider) {
+    return FlagBundle.error(
+      ErrorCode.GENERAL,
+      `OpenFeature provider${
+        config.providerName ? ` "${config.providerName}"` : ''
+      } is not a ConfidenceServerProviderLocal. Register one with OpenFeature.setProviderAndWait — typically in instrumentation.ts.`,
```
nicklasl:
```suggestion
      } is not a ConfidenceServerProviderLocal. Register one with OpenFeature.setProviderAndWait.`,
```

openfeature-provider/js/src/pages-router/server.ts
```diff
@@ -0,0 +1,69 @@
...
+    );
+  }
+
+  const context = typeof config.context === 'function' ? await config.context(ctx) : config.context;
+  const bundle = await provider.resolve(context, config.flags ?? []);
+  // Replace the raw resolveToken in place with its sealed (AES-GCM) form.
+  // The apply API route opens it; the client never sees the raw value.
+  return { ...bundle, resolveToken: sealResolveToken(bundle.resolveToken) };
```
nicklasl, line 61: We should not return the full bundle here. Remove the `resolveToken` so we don't leak it.

openfeature-provider/js/src/pages-router/constants.ts
```diff
@@ -0,0 +1 @@
+export const DEFAULT_APPLY_PATH = '/api/confidence/apply';
```
nicklasl: nit: could move into `common.ts`

openfeature-provider/js/src/pages-router/client.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  const apply = useCallback(
+    async (flagName: string) => {
+      if (!resolveToken) return;
+      await fetch(apiPath, {
+        method: 'POST',
+        headers: { 'content-type': 'application/json' },
+        body: JSON.stringify({ resolveToken, flagName }),
+        keepalive: true,
+      });
```
nicklasl: What are the implications if endpoint returns 503 or is not registered?

openfeature-provider/js/src/pages-router/api.ts
```diff
@@ -0,0 +1,51 @@
...
+ * Builds the POST handler that the client `ConfidencePagesProvider` calls to
+ * fire exposure events. Mount it at `/api/confidence/apply` (or pass a custom
+ * `apiPath` to `ConfidencePagesProvider` if you mount it elsewhere).
+ *
+ * @example
+ * // pages/api/confidence/apply.ts
+ * import { applyHandler } from '@spotify-confidence/openfeature-server-provider-local/pages-router/api';
+ * export default applyHandler();
+ */
+export function applyHandler(opts: ApplyHandlerOptions = {}): NextApiHandler {
```
nicklasl, line 20: **applyHandler ignores bundle.errorCode — App Router doesn't** 

In `react/server.tsx` the server action gates exposure:
```tsx
if (!bundle.errorCode && isConfidenceServerProviderLocal(serverProvider)) {
  serverProvider.applyFlag(bundle.resolveToken, flagName);
}
```

`flag-bundle.ts` shows FlagBundle.error() sets resolveToken: ''. In the Pages Router path, that empty token gets sealed (test confirms sealResolveToken('') round-trips), shipped to the client, posted back, opened to '', and `provider.applyFlag('', flagName)` is called regardless. Behavior diverges from App Router for the error case.

Fix options:
- In applyHandler: `if (!openedToken) { res.status(204).end(); return; }`.
- Or seal errorCode alongside the token so the route can early-return.

openfeature-provider/js/src/pages-router/client.tsx
nicklasl: or what if it's a network issue? promise rejects?

nicklasl: Something like:
```tsx
const res = await fetch(apiPath, { ... });
if (!res.ok && process.env.NODE_ENV !== 'production') {
  console.warn(`[Confidence] apply failed: ${res.status} ${res.statusText}. ` +
    `Mount applyHandler at ${apiPath} and ensure a ConfidenceServerProviderLocal is registered.`);
}
```

openfeature-provider/js/src/pages-router/constants.ts
andreas-karlsson: Good idea — first attempt was to merge into `common.ts`, but that caused a bundle leak: `common.ts` also exports `getConfidenceProvider` which imports `@openfeature/server-sdk`, and tsdown didn't tree-shake the unused export when `client.tsx` imported only the constant from there, so `async_hooks` ended up in the browser bundle (same failure mode as the very first issue we hit when we built this).

Final landing: inlined `const DEFAULT_APPLY_PATH = '/api/confidence/apply'` at the top of `client.tsx` since it's only used in that file. No cross-file import, no leak surface, and `constants.ts` is gone — net same file count as the merge suggestion.

openfeature-provider/js/src/pages-router/client.tsx
andreas-karlsson: Done — fetch wrapped in try/catch, dev-only `console.warn` on both `!res.ok` (using your suggested wording, including the apiPath / "register a ConfidenceServerProviderLocal" hint) and rejected promises. Production stays silent.

openfeature-provider/js/src/pages-router/api.ts
andreas-karlsson: Good catch. Went with the `if (!openedToken) { res.status(204).end(); return; }` early-return after `openResolveToken` — the empty-token-from-`FlagBundle.error()` case is caught without needing to seal any extra metadata alongside the token. Added a test that seals an empty string, posts it, and asserts `applyFlag` is not called and the response is 204. Source comment links to the App Router's `!bundle.errorCode` gate so future readers see the parallel.

nicklasl (approved):
*Jedi Council ignites green saber* 🟢

LGTM — both blocking concerns addressed (errorCode gate in `applyHandler` + dev-warn on apply failures), and the `withConfidence(gssp, opts?)` refactor is genuinely nicer than what I first reviewed. Approving.

---

**One follow-up** — the `Build & Test` failure is the bundle-split sentinel in `Dockerfile:340`. It uses `find ... ! -name <basename>`, which matches **basename only**, so:

- `dist/pages-router/server.js` and `dist/pages-router/client.js` slip through by sheer naming coincidence (the existing `! -name 'server.js' ! -name 'client.js'` entries match them too)
- `dist/pages-router/api.js` has no `api.js` twin elsewhere → caught as "unexpected"

Quick fix is to add `! -name 'api.js'`, but that just papers over the lie. Suggest switching to `! -path` so the allowlist is explicit and the check actually tells the truth:

```dockerfile
RUN set -e; \
    echo "Verifying no bundle splitting in JS artifacts..."; \
    UNEXPECTED_FILES=$(find dist -name '*.js' \
      ! -path 'dist/index.node.js' \
      ! -path 'dist/index.inlined.js' \
      ! -path 'dist/index.fetch.js' \
      ! -path 'dist/server.js' \
      ! -path 'dist/client.js' \
      ! -path 'dist/pages-router/server.js' \
      ! -path 'dist/pages-router/client.js' \
      ! -path 'dist/pages-router/api.js' | head -10); \
    if [ -n "\$UNEXPECTED_FILES" ]; then \
      echo ""; \
      echo "❌ ERROR: Bundle splitting detected!"; \
      echo ""; \
      echo "Found unexpected JavaScript files in dist/:"; \
      echo "\$UNEXPECTED_FILES"; \
      echo ""; \
      echo "Only expected entry point files should be present."; \
      echo "Check tsdown.config.ts configuration to prevent code splitting."; \
      echo ""; \
      exit 1; \
    fi; \
    echo "✅ No bundle splitting detected - only expected files present"
```

Catches actual chunk-splits (which the check was *meant* to find) instead of relying on basename collisions to hide them. 🛠️

nicklasl (approved):

PR #395, opened 2026-05-05
feat(cloudflare): add Wrangler extension hooks to deployer

MattiasMTS:
## Summary
- add `WRANGLER_CONFIG_APPEND_FILE` so deployer users can append opt-in Wrangler TOML such as `tail_consumers` and observability destinations without forking the generated config
- require appended TOML to start with a table header, avoiding accidental top-level keys being parsed inside an existing table
- add `WRANGLER_DEPLOY_TAG` / `WRANGLER_DEPLOY_MESSAGE` for Cloudflare Worker version/deployment labels
- add newline-based `WRANGLER_DEPLOY_ARGS` / `WRANGLER_DEPLOY_ARGS_FILE` as an escape hatch for less common `wrangler deploy` flags
- harden deployer debug output for unset `CARGO_HOME` and clean up ShellCheck warnings in the touched script

## Motivation
The deployer owns the generated `wrangler.toml`, which makes it hard for users to attach account-specific observability pipelines such as Tail Workers or OTEL log destinations. The file-based append hook keeps the deployer closed to provider-specific wiring while allowing users to extend Wrangler config in one predictable way.

## Testing
- `bash -n confidence-cloudflare-resolver/deployer/script.sh`
- `shellcheck confidence-cloudflare-resolver/deployer/script.sh`
- temp smoke test with local resolver state plus stubbed `curl`/`worker-build`/`wrangler`, verifying relative append file resolution, appended `[[tail_consumers]]` config, and `wrangler deploy --tag ... --message ... --minify --keep-vars` arguments

### DIFF
--- a/confidence-cloudflare-resolver/deployer/script.sh
+++ b/confidence-cloudflare-resolver/deployer/script.sh
@@ -14,6 +14,12 @@ CONFIDENCE_CLIENT_SECRET=${CONFIDENCE_CLIENT_SECRET:=}
 NO_DEPLOY=${NO_DEPLOY:=}
 FORCE_DEPLOY=${FORCE_DEPLOY:=}
 WORKER_NAME_PREFIX=${WORKER_NAME_PREFIX:=}
+WRANGLER_CONFIG_APPEND_FILE=${WRANGLER_CONFIG_APPEND_FILE:=}
+WRANGLER_DEPLOY_ARGS=${WRANGLER_DEPLOY_ARGS:=}
+WRANGLER_DEPLOY_ARGS_FILE=${WRANGLER_DEPLOY_ARGS_FILE:=}
+WRANGLER_DEPLOY_TAG=${WRANGLER_DEPLOY_TAG:=}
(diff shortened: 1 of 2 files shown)


confidence-cloudflare-resolver/deployer/script.sh
```diff
@@ -226,6 +231,74 @@ check_file() {
...
+resolve_input_path() {
+    local input_path="$1"
+
+    case "$input_path" in
+        /*) printf '%s\n' "$input_path" ;;
+        *) printf '%s/%s\n' "$INITIAL_WORKDIR" "$input_path" ;;
+    esac
+}
+
+validate_wrangler_config_append() {
```
MattiasMTS, line 243: not sure if we want this really - wrangler should handle it tbh.

fabriziodemaria: I'd be ok to remove it, if you want to push the extra commit. Not a blocker though

fabriziodemaria (approved):

PR #400, opened 2026-05-07
feat(cloudflare): add telemetry collection and /metrics endpoint

vahidlazio:
## Summary

Add telemetry collection to the Cloudflare resolver with the **same metric names** as all other providers, enabling shared Grafana dashboards.

### Key discovery: `scheduler.wait(0)` unfreezes CF Workers timers

CF Workers freeze `Date.now()` and `performance.now()` during synchronous CPU work (Spectre mitigation), making internal latency measurement appear impossible. We discovered that **`scheduler.wait(0)`** — a zero-delay yield to the runtime — unfreezes the clock with no measurable overhead. This enables inline CPU time measurement identical to other providers, without any external API dependency.

### What's included

- **Inline resolve latency** via `scheduler.wait(0)` + `Date.now()` (1ms resolution, Grafana interpolates to sub-ms)
- **Resolve rate telemetry** by reason (MATCH, NO_SEGMENT_MATCH, ERROR, etc.)
- **`/metrics` endpoint** serving Prometheus exposition format (`confidence_resolve_latency_microseconds` histogram + `confidence_resolves_total` counters)
- **Queue-based aggregation** — per-isolate deltas accumulated into KV-backed cumulative snapshot
- **Backend telemetry** — `WriteFlagLogsRequest` with SDK ID (`SDK_ID_CLOUDFLARE_RESOLVER = 25`), resolver version, resolve rates, and latency histogram
- **`Cache-Control: no-store`** on `/metrics` response
- **Deployer** auto-creates KV namespace for metrics storage

### Data flow

```
Hot path:     Date.now() → resolve → scheduler.wait(0) → Date.now() → elapsed_us
              Push {elapsed_us, reasons} to PENDING_METRICS
wait_until:   Drain metrics → TELEMETRY.record_latency_us / mark_resolve
              → checkpoint() → Queue
Queue consumer (batched, all isolates):
  ├─ KV: read cumulative → accumulate deltas → write snapshot + prometheus text
  └─ POST aggregated WriteFlagLogsRequest to Confidence backend
/metrics:     Read prometheus text from KV
```

### What was explored and removed

The initial approach sourced latency from Cloudflare's GraphQL Analytics API (`workersInvocationsAdaptive` percentiles). This worked but added complexity: cursor-based pagination, rate-limiting, percentile-to-histogram distribution, cached percentile estimation. Once `scheduler.wait(0)` was proven to unfreeze timers, the entire GraphQL machinery was removed (~360 lines), leaving a clean inline timer identical to other providers.

## Test plan

- [x] `cargo check -p confidence-cloudflare-resolver` compiles clean
- [x] `cargo test -p confidence_resolver -- telemetry` passes (25 tests)
- [x] Deploy to test account and verify `/metrics` returns valid Prometheus text
- [x] Verify resolve reasons accumulate correctly (MATCH, NO_SEGMENT_MATCH, ERROR)
- [x] Verify latency histogram matches warm resolve times (~1-2ms p50)
- [x] Verify telemetry appears on Grafana dashboard (resolve rate continuous, latency continuous)
- [x] Verify `scheduler.wait(0)` adds no measurable latency to resolve path (0ms overhead in 19/20 samples)
- [x] Verify metric names match other providers (`confidence_resolve_latency_microseconds`, `confidence_resolves_total`)


confidence-cloudflare-resolver/src/lib.rs
```diff
@@ -181,6 +210,7 @@ pub async fn main(req: Request, env: Env, ctx: Context) -> Result<Response> {
                             .evaluation_context
                             .clone()
                             .unwrap_or_default();
+                        let start = js_sys::Date::now();
```
andreas-karlsson: Prefer using Performance::now() for better resolution.

confidence-cloudflare-resolver/src/lib.rs
```diff
@@ -142,6 +158,19 @@ pub async fn main(req: Request, env: Env, ctx: Context) -> Result<Response> {
...
+        .get_async("/metrics", |_req, ctx| {
+            let allowed_origin = allowed_origin_env.clone();
+            async move {
+                let text = match ctx.env.kv("METRICS_KV") {
+                    Ok(kv) => kv.get("prometheus").text().await.unwrap_or(None),
+                    Err(_) => None,
+                };
+                let body = text.unwrap_or_default();
+                let headers = Headers::new();
+                headers.set("Content-Type", "text/plain; version=0.0.4; charset=utf-8")?;
```
andreas-karlsson: I think there might be some special Content-Type we should use for Prometheus.

confidence-cloudflare-resolver/src/lib.rs
```diff
@@ -142,6 +158,19 @@ pub async fn main(req: Request, env: Env, ctx: Context) -> Result<Response> {
     let router = Router::new();
 
     let response = router
+        .get_async("/metrics", |_req, ctx| {
```
andreas-karlsson: We should use this [parser](https://github.com/spotify/confidence-resolver/blob/3d306834936d2f89fd346579c3bd183480a613ed/openfeature-provider/go/confidence/integration_test.go#L750) to check that we can parse the output of this.

Edit: No need to do this since it's the same serializer that we test in go.

confidence-cloudflare-resolver/deployer/script.sh
```diff
@@ -366,6 +366,55 @@ else
...
+else
+    echo "✅ KV namespace '$KV_NAMESPACE_TITLE' already exists (id: $KV_NAMESPACE_ID)"
+fi
+
+# Append KV binding to wrangler.toml if namespace was created
+if [ -n "$KV_NAMESPACE_ID" ]; then
+    cat >> wrangler.toml <<EOF
+
+[[kv_namespaces]]
+binding = "METRICS_KV"
```
andreas-karlsson: Change to CONFIDENCE_METRICS_KV

confidence-cloudflare-resolver/src/lib.rs
andreas-karlsson: Also we could consider supporting the same options on this endpoint like we do in local: https://github.com/spotify/confidence-resolver/blob/3d306834936d2f89fd346579c3bd183480a613ed/openfeature-provider/go/confidence/provider.go#L356

But that could also wait for a followup PR. Maybe even better.

confidence-cloudflare-resolver/src/lib.rs
vahidlazio: Already correct — `text/plain; version=0.0.4` is the standard Prometheus exposition format Content-Type.

confidence-cloudflare-resolver/src/lib.rs
```diff
@@ -141,6 +157,20 @@ pub async fn main(req: Request, env: Env, ctx: Context) -> Result<Response> {
...
+            let allowed_origin = allowed_origin_env.clone();
+            async move {
+                let text = match ctx.env.kv("CONFIDENCE_METRICS_KV") {
+                    Ok(kv) => kv.get("prometheus").text().await.unwrap_or(None),
+                    Err(_) => None,
+                };
+                let body = text.unwrap_or_default();
+                let headers = Headers::new();
+                headers.set("Content-Type", "text/plain; version=0.0.4; charset=utf-8")?;
```
nicklasl: what version is this?

nicklasl: maybe break out to constant so we can name it?

confidence-cloudflare-resolver/src/lib.rs
```diff
@@ -249,8 +322,19 @@ pub async fn main(req: Request, env: Env, ctx: Context) -> Result<Response> {
...
         .await;
 
-    // Use ctx.waitUntil to run logging after response is returned
+    // Use ctx.waitUntil to run logging and telemetry after response is returned.
     ctx.wait_until(async move {
+        PENDING_METRICS.with(|m| {
+            for metrics in m.borrow_mut().drain(..) {
+                if metrics.elapsed_us > 0 {
+                    TELEMETRY.record_latency_us(metrics.elapsed_us.min(u32::MAX as u64) as u32);
+                }
```
nicklasl: This will effectively filter out ALL sub milli second values. Since Date.now() only has milli precision. It would mean that we only return metrics for resolve latency > 1 ms.

confidence-cloudflare-resolver/src/lib.rs
```diff
@@ -194,24 +229,62 @@ pub async fn main(req: Request, env: Env, ctx: Context) -> Result<Response> {
...
+                                ).unwrap_or(wasm_bindgen::JsValue::UNDEFINED);
+                                if let Ok(promise) = js_sys::Function::from(wait)
+                                    .call1(&scheduler, &wasm_bindgen::JsValue::from(0))
+                                {
+                                    let _ = wasm_bindgen_futures::JsFuture::from(
+                                        js_sys::Promise::from(promise)
+                                    ).await;
+                                }
                             }
                         }
```
nicklasl: what happens if this is changed on the platform?
is there any way we can guard against panics or something?

confidence-cloudflare-resolver/src/lib.rs
```diff
@@ -141,6 +157,20 @@ pub async fn main(req: Request, env: Env, ctx: Context) -> Result<Response> {
     let router = Router::new();
 
     let response = router
+        .get_async("/metrics", |_req, ctx| {
```
nicklasl: remind me, are these workers running public? so anyone can call this endpoint and get data (no auth).

vahidlazio: added client secret guard.

confidence-cloudflare-resolver/src/lib.rs
vahidlazio: it was falling to 0 .. changed to optional and this to return none so we don't report.

confidence-cloudflare-resolver/src/lib.rs
vahidlazio: fixed.

confidence-cloudflare-resolver/src/lib.rs
vahidlazio: this is prometheus exposition format version — the only version that exists. Every /metrics endpoint uses 0.0.4. asked ai if we are required to 

`
 No. Prometheus scrapers accept text/plain without the version parameter. But including it is the convention — every /metrics endpoint does it, and it's what the https://prometheus.io/docs/instrumenting/exposition_formats/#text-based-format documents. Harmless to keep, confusing to explain if removed.  
`

nicklasl (approved):

confidence-cloudflare-resolver/deployer/README.md
```diff
@@ -120,6 +121,33 @@ When integrating with the Cloudflare resolver, you have two options:
...
+
+Returns Prometheus exposition format with:
+- `confidence_resolve_latency_microseconds` — histogram (sum, count, cumulative `le` buckets)
+- `confidence_resolves_total` — counter by resolve reason
+
+Metrics are accumulated in a [KV namespace](https://developers.cloudflare.com/kv/platform/pricing/) (`CONFIDENCE_METRICS_KV`). Set `ENABLE_METRICS` to have the deployer create the KV namespace and bind it to the Worker. Without it, the `/metrics` endpoint returns empty and no KV writes occur.
+
+### Backend telemetry
+
+Resolve rates and latency are always sent to the Confidence backend via `WriteFlagLogsRequest`, regardless of the `DISABLE_METRICS` setting. The `/metrics` endpoint and KV store are only needed for direct Prometheus scraping — backend telemetry flows through the queue consumer independently.
```
nicklasl:
```suggestion
Resolve rates and latency are always sent to the Confidence backend via `WriteFlagLogsRequest`, regardless of the `ENABLE_METRICS` setting. The `/metrics` endpoint and KV store are only needed for direct Prometheus scraping — backend telemetry flows through the queue consumer independently.
```

PR #420, opened 2026-05-27
feat(go): skip apply via context key

MattiasMTS:
## Summary
- add `confidence.SkipApplyContextKey` for Go `context.Context` so callers can force resolve requests to use `apply=false`
- apply the context-key behavior to both OpenFeature evaluation methods and direct `provider.Resolve(ctx, ...)`
- add regression coverage for typed object evaluation and direct `Resolve` with the skip-apply context key

## How it works
Set `confidence.SkipApplyContextKey` to `true` on the Go `context.Context` passed into the provider. The provider reads that key before building the resolver request and flips `apply` to `false`. The key is not part of the OpenFeature evaluation context, so it is not sent to the resolver as a targeting attribute.

```go
ctx := context.WithValue(
    context.Background(),
    confidence.SkipApplyContextKey,
    true,
)

resp, err := provider.Resolve(
    ctx,
    openfeature.FlattenedContext{
        "targeting_key": "user-123",
    },
    []string{"my-flag"},
    true, // normally apply, but the context key forces apply=false
)
```

The same context key also works with normal OpenFeature evaluation calls:

```go
ctx := context.WithValue(context.Background(), confidence.SkipApplyContextKey, true)

result := provider.BooleanEvaluation(
    ctx,
    "my-flag",
    false,
    openfeature.FlattenedContext{"targeting_key": "user-123"},
)
```

When skip-apply is enabled, the resolve response can include a resolve token for deferred apply. The caller can later record exposure with `ApplyFlags`.

## Testing
- `go test ./confidence -run TestLocalResolverProvider -count=1`
- `go test ./confidence -run 'TestLocalResolverProvider_(ObjectEvaluationWithSkipApplyContextReturnsTypedStruct|ResolveWithSkipApplyContext)|TestProcessTargetingKey|TestFlattenedContextToProto' -count=1`
- `git diff --check`

Note: full `go test ./confidence` still hits existing integration-style tests that require configured client secret/backend state.

### DIFF
--- a/openfeature-provider/go/confidence/provider.go
+++ b/openfeature-provider/go/confidence/provider.go
@@ -19,9 +19,14 @@ import (
 	"google.golang.org/protobuf/types/known/structpb"
 )
 
+type contextKey string
+
 const (
 	defaultStatePollIntervalSeconds = 10
 	defaultLogPollIntervalSeconds   = 15
+
+	// SkipApplyContextKey disables apply-on-resolve when set to true in context.Context.
+	SkipApplyContextKey contextKey = "_confidence_skip_apply"
 )
 
 type LocalResolverSupplier func(context.Context, lr.LogSink) lr.LocalResolver
@@ -248,7 +253,7 @@ func evaluate[T any](
 		})
 	}
 
-	response, err := p.resolveFlags(evalCtx, []string{requestFlagName}, true)
+	response, err := p.resolveFlags(evalCtx, []string{requestFlagName}, !skipApplyFromContext(ctx))
 	if err != nil {
 		var matErr *MaterializationNotSupportedError
 		if errors.As(err, &matErr) {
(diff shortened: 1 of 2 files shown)


MattiasMTS:
Raising here as well, let me know if you want function signature to be:

```
func (p *LocalResolverProvider) ObjectEvaluationWithOptions(
        ctx context.Context,
        flag string,
        defaultValue any,
        evalCtx openfeature.FlattenedContext,
        opts ...EvaluationOption,
  ) openfeature.InterfaceResolutionDetail {
        options := defaultEvaluationOptions()
        for _, opt := range opts {
                opt(&options)
        }

        return evaluate(p, ctx, flag, defaultValue, evalCtx, options.apply)
  }
```
in the future instead -> allowing more flexibility (avoid breaking change etc)

MattiasMTS:
@nicklasl

PR #421, opened 2026-05-27
feat: support skipping apply using context key

nicklasl:
## Summary

- Adds support for a reserved evaluation context key `_confidence_skip_apply` across JS, Java, Go, Python, and Rust OpenFeature providers
- When set to `true`, the provider resolves the flag with `apply=false`, suppressing the exposure event
- The key is stripped from the context before it reaches the resolver — it never leaks into the targeting context

## Test plan

- [x] JS: vitest passes — verifies `resolveProcess` called with `apply: false` and key stripped from context
- [x] Java: `OpenFeatureUtilsTest` — `isSkipApply` returns correct values, `convertToProto` strips the key
- [x] Go: `TestLocalResolverProvider_SkipApplyContextKey` — resolves correctly with key in context
- [x] Python: `TestSkipApply` — resolves with skip apply, key stripped from attributes
- [x] Rust: unit tests for key stripping and default behavior

Replaces https://github.com/spotify/confidence-resolver/pull/420

### DIFF
--- a/openfeature-provider/js/src/ConfidenceServerProviderLocal.ts
+++ b/openfeature-provider/js/src/ConfidenceServerProviderLocal.ts
@@ -200,8 +200,9 @@ export class ConfidenceServerProviderLocal implements Provider {
     const startMs = performance.now();
     try {
       const [flagName] = flagKey.split('.', 1);
-      const skipApply = context._confidence_skip_apply === true;
-      const resolution = await this.resolve(context, [flagName], !skipApply);
+      const { _confidence_skip_apply, ...cleanContext } = context;
+      const skipApply = _confidence_skip_apply === true;
+      const resolution = await this.resolve(cleanContext as EvaluationContext, [flagName], !skipApply);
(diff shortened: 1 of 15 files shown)


openfeature-provider/js/src/ConfidenceServerProviderLocal.ts
```diff
@@ -200,7 +200,8 @@ export class ConfidenceServerProviderLocal implements Provider {
     const startMs = performance.now();
     try {
       const [flagName] = flagKey.split('.', 1);
-      const resolution = await this.resolve(context, [flagName], true);
+      const skipApply = context._confidence_skip_apply === true;
```
andreas-karlsson: Here we don't remove it from the context it? I think we did that in Go.

nicklasl: good catch! yes, the idea is that we should exclude it once we read it.

andreas-karlsson (approved):

PR #423, opened 2026-05-28
fix(resolver): handle diamond and circular segment dependencies

nicklasl:
## Summary

- **Diamond segment dependencies** (A → B → C, A → C) were falsely detected as circular, causing `RESOLVE_REASON_ERROR` for affected flags. Fixed by cloning the visited set per branch so siblings don't pollute each other.
- **Circular segment dependencies** (A → B → A) caused an internal error that failed the entire resolve. Aligned with the Java resolver: return `false` (no match) instead of erroring.

## Test plan

- [x] Added `diamond_segment_dependency` and `diamond_unaffected_flag` to resolver spec tests (shared with Java resolver)
- [x] Added programmatic tests covering single-flag, batch, and unaffected-flag scenarios
- [x] Existing `circular_segment_dependency` and `self_referencing_segment` spec tests updated to expect `NO_SEGMENT_MATCH` instead of `InternalError`
- [x] All 362 lib tests pass
- [x] Java resolver passes the same spec tests


andreas-karlsson (approved):
Nice!

PR #426, opened 2026-05-29
fix(java): drop pending poll on local provider shutdown

mfranberg:
## Problem

In production the local provider logs this on essentially **every** pod termination:

```
com.spotify.confidence.sdk.OpenFeatureLocalResolveProvider: Flags fetcher executor did not terminate gracefully
```

Root cause: the flags-fetcher runs a poll task that reschedules itself, so there is always a poll pending in the future. `Executors.newScheduledThreadPool` returns a `ScheduledThreadPoolExecutor` whose `executeExistingDelayedTasksAfterShutdownPolicy` defaults to **true**, so `shutdown()` keeps that pending task and `awaitTermination` waits for a poll that won't fire for another ~15s. It times out and force-kills the executor — every time, not just when a poll is genuinely in flight.

## Fix

- Set `executeExistingDelayedTasksAfterShutdownPolicy(false)` on both schedulers, so `shutdown()` drops the pending poll and `awaitTermination` returns as soon as any actually-running task finishes.
- Bump the grace period from 1s to 5s so a poll caught genuinely mid-fetch (blocking CDN GET) or mid-state-push (WASM update across the pool) still gets a chance to finish before being forced.

Logs are unaffected — they're still flushed by `resolver.close()` / `flagLogger.shutdown()` after the executors stop.

## Test

`OpenFeatureLocalResolveProviderShutdownTest` initializes the provider (which schedules a poll) and immediately shuts down, asserting the fetcher terminates without being force-killed (`forcedFetcherShutdown == false`). This fails on the old behavior (the pending poll forces the timeout) and passes with the policy fix.

> The bundled `confidence_resolver.wasm` is git-ignored and built locally, so I can't run the resolver tests in my environment (no `wasm32` rust toolchain); they need CI. The change compiles.

### DIFF
--- a/openfeature-provider/java/src/main/java/com/spotify/confidence/sdk/OpenFeatureLocalResolveProvider.java
+++ b/openfeature-provider/java/src/main/java/com/spotify/confidence/sdk/OpenFeatureLocalResolveProvider.java
@@ -57,8 +57,8 @@ public class OpenFeatureLocalResolveProvider implements FeatureProvider {
   private final MaterializationStore materializationStore;
   private static final Duration ASSIGN_LOG_FLUSH_INTERVAL = Duration.ofMillis(100);
   private static final Duration DEFAULT_POLL_INTERVAL = Duration.ofSeconds(15);
-  private final ScheduledExecutorService flagsFetcherExecutor =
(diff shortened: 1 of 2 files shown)


nicklasl:
1 second should be enough for everyone! 🧌

nicklasl (approved):

PR #429, opened 2026-05-30
perf(local_resolver): cache exported WASM function handles to avoid per-call call-engine allocation

avoxelisnotapixel:
Evaluating a flag looked up a WASM function from scratch every time, which allocated a new execution stack.
This causes a *ton* of unnecessary allocations.

Caching the function handles reduces allocations to almost nothing.


nicklasl:
Thanks for the contribution, I picked this up in https://github.com/spotify/confidence-resolver/pull/430

PR #431, opened 2026-06-01
fix: register telemetry on bundle resolve path

nicklasl:
## Summary

- The `resolve()` method (used by Next.js pages-router and React Server Components for flagbundle resolves) bypassed `evaluate()`, which was the only place calling `registerResolve()`. This meant telemetry (resolve rate, latency) was never emitted for bundle resolves, while resolve logs and apply logs worked fine.
- Adds `RESOLVE_REASON_BUNDLE` to the proto enum and calls `registerResolve()` from the bundle resolve path in JS, Java, and Go providers.
- Extracts a private `resolveFlags()` in JS so `evaluate()` and `resolve()` each register telemetry independently without double-counting.

## Test plan

- [x] JS integration test (`resolve-logging.test.ts`) verifies telemetry data (resolve rate + latency) appears in the network payload after a `provider.resolve()` call
- [x] All existing JS tests pass (`npx vitest run --exclude='**/*.e2e.test.ts'`) — 153 passed, 8 skipped (pre-existing `flaglogs` suite requires mock server)
- [x] Go compiles cleanly (`go build ./...`)
- [x] Java: verified `doRegisterResolve` is called via `whenComplete` on the resolve `CompletionStage` — correct placement after `thenApply`, proper error/success reason mapping, no double-counting with the `evaluate()` path


fabriziodemaria (commented):
## Python: `_confidence_skip_apply` extraction mutates the caller's context

`provider.py:459` uses `evaluation_context.attributes.pop(...)` which mutates the dict in-place. If a caller reuses the same `EvaluationContext` across multiple evaluations, the first call strips the key and subsequent calls will send `apply=True` even though the caller intended to skip it:

```python
ctx = EvaluationContext(
    targeting_key="user-123",
    attributes={
        "visitor_id": "tutorial_visitor",
        "_confidence_skip_apply": True,
    },
)

# First call: pop() removes the key → skip_apply=True, apply=False ✓
result1 = provider.resolve_string_details("flag.key", "default", ctx)

# Second call: key is already gone → skip_apply=False, apply=True ✗
result2 = provider.resolve_string_details("flag.key", "default", ctx)
```

Every other provider avoids this: JS uses destructuring, Go deletes from an already-copied map, Java reads then filters in `convertToProto`, Rust calls `.remove()` on a clone.

Suggest replacing the `pop` with a non-mutating read:

```python
skip_apply = (
    evaluation_context.attributes.get("_confidence_skip_apply", False)
    is True
)
```

and filtering the key out in `_context_to_proto` instead (similar to the Java approach), or working on a shallow copy of `attributes`.

fabriziodemaria:
**Correction on my earlier review:** The comment about Python's `_confidence_skip_apply` / `pop()` mutation does not apply to this PR. That change was introduced in PR #421 (commit 0ad63cc). This PR (#431) only touches JS, Java, and Go for bundle resolve telemetry. Apologies for the noise.

openfeature-provider/go/confidence/provider.go
```diff
@@ -407,7 +407,19 @@ func (p *LocalResolverProvider) Resolve(
 		requestFlags[i] = "flags/" + name
 	}
 
-	return p.resolveFlags(evalCtx, requestFlags, apply)
+	start := time.Now()
+	resp, err := p.resolveFlags(evalCtx, requestFlags, apply)
+
+	reason := resolver.ResolveReason_RESOLVE_REASON_BUNDLE
+	if err != nil {
+		reason = resolver.ResolveReason_RESOLVE_REASON_ERROR
```
fabriziodemaria, line 415: Too bad we can't distinguish errors between single and bundle, perhaps worth a new enum?

fabriziodemaria (approved):

PR #432, opened 2026-06-04
fix(resolver): treat empty resolve token as no-op in apply_flags

MattiasMTS:
## Problem

The Cloudflare edge resolver forces `apply = true` on `flags:resolve` (`confidence-cloudflare-resolver/src/lib.rs`) and therefore returns no `resolveToken`. However, `@spotify-confidence/sdk` always resolves with `apply: false` and then issues a background `flags:apply` (there is no option to disable it), carrying an **empty** `resolve_token`.

`AccountResolver::apply_flags` calls `decrypt_resolve_token(&request.resolve_token)` unconditionally; on an empty buffer the decrypt reads bytes `0..16`, fails, and the Cloudflare handler returns **500**.

Following the [documented Cloudflare edge integration](https://confidence.spotify.com/docs/sdks/edge/cloudflare) verbatim (backend SDK + service binding), this returns 500 on essentially every background apply — millions per hour in our deployment. Exposures are unaffected (edge resolvers record apply data at resolve time, per the [apply-event docs](https://confidence.spotify.com/docs/sdks/apply-event)), so it is wasted invocations and error noise rather than data loss — but it is very high volume.

## Fix

An apply request with an empty `resolve_token` has no assignments to apply, so `apply_flags` returns `Ok(())` early instead of attempting to decrypt an empty buffer. Non-empty tokens are unchanged — a malformed/undecryptable token still errors.

## Test

Added `apply_flags_empty_resolve_token_is_noop`, which builds a resolver and asserts `apply_flags` returns `Ok(())` for an `ApplyFlagsRequest` with an empty `resolve_token`.

### DIFF
--- a/confidence-resolver/src/lib.rs
+++ b/confidence-resolver/src/lib.rs
@@ -937,6 +937,13 @@ impl<'a, H: Host> AccountResolver<'a, H> {
     }
 
     pub fn apply_flags(&self, request: &flags_resolver::ApplyFlagsRequest) -> Result<(), String> {
+        // The Cloudflare edge resolver applies at resolve time (apply=true) and
+        // returns no resolve token, but SDKs still issue a background apply with
+        // an empty token. There is nothing to apply, so no-op instead of failing
+        // to decrypt an empty buffer.
+        if request.resolve_token.is_empty() {
+            return Ok(());
+        }
         let send_time_ts = request.send_time.as_ref().ok_or("send_time is required")?;
         let send_time = to_date_time_utc(send_time_ts).ok_or("invalid send_time")?;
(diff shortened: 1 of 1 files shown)


nicklasl:
Thanks for the investigation and clear write-up! The root cause analysis is spot on.

However, since this is a Cloudflare-specific issue (the CF resolver forces `apply=true` and drops the resolve token), I'd prefer to keep the fix in the CF harness rather than the core resolver — that way we don't change behavior for all providers for a problem only one of them creates.

I've put up an alternative in #433 that handles this in `confidence-cloudflare-resolver` by short-circuiting the `flags:apply` handler when the token is empty, before even constructing a resolver.

PR #434, opened 2026-06-05
fix(cloudflare): no-op apply and telemetry for service binding SDKs

nicklasl:
## Summary

- Short-circuit `flags:apply` with 200 when `resolve_token` is empty — the resolver already applies at resolve time (`apply=true`), so the background apply SDKs send is redundant and was causing 500s when trying to decrypt an empty token
- Add no-op `telemetry:upload` route — SDKs using service bindings route all requests through the resolver, including telemetry which has no route (404)
- Add service binding integration example with `waitUntil` to deployer README

Closes #432

## Test plan

- [x] Deployed fixed resolver to Cloudflare test account
- [x] Verified `flags:apply` with empty token returns 200
- [x] Verified `telemetry:upload` returns 200
- [x] Verified full SDK flow (resolve + apply) via service binding returns all 200s
- [x] Verified resolve still works correctly (`MATCH` with expected variant)

### DIFF
--- a/confidence-cloudflare-resolver/src/lib.rs
+++ b/confidence-cloudflare-resolver/src/lib.rs
@@ -331,6 +331,14 @@ pub async fn main(req: Request, env: Env, ctx: Context) -> Result<Response> {
                             }
                         };
 
+                        // This resolver forces apply=true at resolve time and
+                        // returns no resolve token. Some SDKs still send a background
+                        // apply with an empty token — nothing to do.
+                        if apply_flag_req.resolve_token.is_empty() {
+                            return Response::from_json(&ApplyFlagsResponse::default())?
+                                .with_cors_headers(&allowed_origin);
+                        }
+
                         match state.get_resolver::<H>(
                             &apply_flag_req.client_secret,
                             Struct::default(),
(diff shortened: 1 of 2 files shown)


confidence-cloudflare-resolver/src/lib.rs
```diff
@@ -347,6 +355,9 @@ pub async fn main(req: Request, env: Env, ctx: Context) -> Result<Response> {
                             }
                         }
                     }
+                    "telemetry:upload" => {
+                        Response::ok("")?.with_cors_headers(&allowed_origin)
+                    }
```
fabriziodemaria, line 360: We some telemetry support in CloudFlare now, but I suppose forwarding this data into that is not a simple plug and play?

fabriziodemaria (approved):
A bit sad that no test broke, would be nice to add one for the empty token scenario. Not a blocker

confidence-cloudflare-resolver/src/lib.rs
nicklasl: Yeah, I don't think it is very trivial. Right now I'm struggling how we would like to model it and tell it apart from the other telemetry for the same provider..

PR #436, opened 2026-06-05
fix(python): avoid mutating caller context for skip-apply

fabriziodemaria:
## Summary

- `ConfidenceProvider._resolve_object` extracted `_confidence_skip_apply` with `evaluation_context.attributes.pop(...)`, which **mutates the caller's `EvaluationContext` in place**. If a caller reuses the same context across multiple evaluations, the key is stripped on the first call, so subsequent calls see it missing and send `apply=True` — silently breaking apply suppression after the first resolve.
- Switched to a non-mutating `attributes.get(...)` read, and filter the internal `_confidence_skip_apply` key out when building the proto context (in `_context_to_proto`) so it still never reaches the resolver.
- This matches what every other provider already does (JS destructures, Go deletes from a copied map, Java reads then filters in `convertToProto`, Rust `.remove()`s a clone).

Originally surfaced in review of #431, but the bug was introduced in #421 (commit `0ad63cc`) and only affects the Python provider, so it's fixed here separately.

## Test plan

- Replaced `test_skip_apply_key_is_stripped_from_context` (which asserted the buggy mutation) with `test_skip_apply_does_not_mutate_caller_context`, which resolves twice with the same context and asserts the key remains intact across calls.
- `python3 -m py_compile` passes on both edited files.
- Note: full `make test` requires building the WASM artifact (no local Rust toolchain in this environment); CI will exercise the suite.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/openfeature-provider/python/src/confidence/provider.py
+++ b/openfeature-provider/python/src/confidence/provider.py
@@ -456,7 +456,7 @@ class ConfidenceProvider(AbstractProvider):
             skip_apply = False
             if evaluation_context and evaluation_context.attributes:
                 skip_apply = (
-                    evaluation_context.attributes.pop("_confidence_skip_apply", False)
+                    evaluation_context.attributes.get("_confidence_skip_apply", False)
                     is True
                 )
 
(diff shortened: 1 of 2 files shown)


nicklasl (approved):

nicklasl:
This PR fixes the test issue: https://github.com/spotify/confidence-resolver/pull/437

nicklasl:
please rebase on main instead.

PR #443, opened 2026-06-10
feat(resolve): add support for resolving materialized segments via bloom filters

nicklasl:
## Summary

- Add `PackedBloomFilter` proto message to `ResolverState` (field 10)
- New `bloom_filter.rs` module with Guava MURMUR128_MITZ_64 compatible membership test
- Bloom filter fast path in `targeting_match`: when a bloom filter exists for a materialized segment, resolve locally without suspend/resume
- Falls back to existing materialization context when no bloom filter is present

## Test plan

- [x] Unit tests: all-set, all-clear, unsupported strategy, programmatic insert+query
- [x] Cross-language conformance: real Guava 33.5.0 generated bloom filter verified in Rust
- [x] Resolver spec tests: `bloom_filter_match` and `bloom_filter_no_match`
- [x] Existing materialization tests unaffected (368→369 tests, all green)
- [x] Clippy clean


confidence-resolver/src/bloom_filter.rs
```diff
@@ -0,0 +1,177 @@
...
+}
+
+impl BloomFilter {
+    pub fn from_proto(packed: &BloomFilterPb) -> Fallible<Self> {
+        if packed.strategy != STRATEGY_MURMUR128_MITZ_64 {
+            return Err(crate::err::ErrorCode::from_tag(
+                "bloom_filter.unsupported_strategy",
+            ));
+        }
+        let num_bits = packed.bit_count as u64;
```
fabriziodemaria, line 23: I don't know if it's even possible, but if `bit_count` comes in as 0 or negative we would have either a panic or a wrong evaluation... Maybe good to fail loudly in both cases with clear logs for the user?

confidence-resolver/src/bloom_filter.rs
```diff
@@ -0,0 +1,177 @@
...
+        let num_hash_functions = packed.hash_function_count as u32;
```
fabriziodemaria, line 24: Same here, I guess: it's good to protect against negative packed values that would wrap to huge u32 numbers

confidence-resolver/src/bloom_filter.rs
```diff
@@ -0,0 +1,177 @@
...
+        let mut data = packed.data.clone();
+        // Guava serializes its long[] array in big-endian byte order (DataOutputStream.writeLong).
+        // BitVec<u8, Lsb0> expects bit 0 at the LSB of byte 0 (little-endian within each long).
+        // Reverse each 8-byte chunk to convert from BE to LE so bit indices align.
+        for chunk in data.chunks_exact_mut(8) {
+            chunk.reverse();
+        }
```
fabriziodemaria, line 31: Guava always writes complete longs, so data.len() % 8 == 0 should always hold. But if a malformed proto has e.g. 13 bytes, the last 5 bytes wouldn't be swapped, silently producing wrong bit layout.

confidence-resolver/src/bloom_filter.rs
```diff
@@ -0,0 +1,177 @@
...
+            bit_count,
+            strategy: STRATEGY_MURMUR128_MITZ_64,
+        }
+    }
+
+    #[test]
+    fn all_bits_set_always_matches() {
+        let data = vec![0xFF; 16];
+        let packed = make_packed(&data, 3, 128);
+        let bf = BloomFilter::from_proto(&packed).unwrap();
```
fabriziodemaria: nit: from_proto takes &BloomFilterPb but the data is only used here — taking ownership would avoid cloning the byte buffer and let BitVec::from_vec reuse the allocation (3x → 1x copies). (by AI)

fabriziodemaria (approved):
Seems good, AI mostly pointed out concerns about malformed bloom filter data causing crashes or unwanted resolves, but if we are confident to not produce bad state I think those concerns can be considered minor

PR #445, opened 2026-06-15
feat(js): publish flag evaluations to window global for session recording

nicklasl:
## Summary
- Adds `publishFlagEvaluation` that writes to `window.__confidence.flags` on MATCH evaluations, enabling the session recorder to observe flag assignments and emit `csr:flagEvaluation` plugin events.
- Mirrors the pattern from [confidence-sdk-js#369](https://github.com/spotify/confidence-sdk-js/pull/369).

## Test plan
- [x] Unit tests for `publishFlagEvaluation` (SSR safety, dedup, object preservation)
- [x] Existing test suites pass (59 tests)
- [x] TypeScript type check clean
- [x] Verified end-to-end in a local Next.js app: evaluated `hawkflag` and `tutorial-flag` via the WASM resolver, hydrated results to `window.__confidence.flags` on the client, and confirmed a Proxy-based mock session recorder observed the writes as `csr:flagEvaluation` events. Also tested lazy mount — hid one flag behind a button to verify the global is only written when the component mounts.

### DIFF
--- a/openfeature-provider/js/src/flag-evaluation-global.ts
+++ b/openfeature-provider/js/src/flag-evaluation-global.ts
@@ -1,15 +1,5 @@
-interface ConfidenceFlagEntry {
-  variant: string;
-}
-
 interface ConfidenceGlobal {
-  flags?: Record<string, ConfidenceFlagEntry>;
-}
-
-declare global {
-  interface Window {
-    __confidence?: ConfidenceGlobal;
-  }
+  flags?: Record<string, { variant: string }>;
 }
 
 export function publishFlagEvaluation(name: string, variant: string): void {
--- a/openfeature-provider/js/src/flag-evaluation-global.ts
+++ b/openfeature-provider/js/src/flag-evaluation-global.ts
@@ -1,12 +1,22 @@
-declare const window: { __confidence?: { flags?: Record<string, { variant: string }> } } | undefined;
+interface ConfidenceFlagEntry {
+  variant: string;
+}
+
+interface ConfidenceGlobal {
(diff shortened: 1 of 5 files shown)


andreas-karlsson (commented):
I'm surprised to see this approach of putting the logic on the ConfidenceServerProviderLocal? It normally won't be used in the same JS engine as the session-recorder. I think we should consider moving this functionality to the react hooks.

openfeature-provider/js/src/flag-evaluation-global.ts
```diff
@@ -0,0 +1,22 @@
...
+
+interface ConfidenceGlobal {
+  flags?: Record<string, ConfidenceFlagEntry>;
+}
+
+declare global {
+  interface Window {
+    __confidence?: ConfidenceGlobal;
+  }
+}
```
andreas-karlsson: Do we need to declare this?

openfeature-provider/js/src/react/client.tsx
```diff
@@ -169,6 +170,12 @@ export function useFlagDetails<T extends FlagValue>(
     }
   }, [autoExpose, doExpose]);
 
+  useEffect(() => {
+    if (resolution.reason === 'MATCH' && resolution.variant) {
+      publishFlagEvaluation(`flags/${baseFlagName}`, resolution.variant);
```
andreas-karlsson: Prefer to do this in `doExpose`? And I think we should do it irrespectively of shouldApply, but only if we have a variant.

fabriziodemaria (approved):

PR #454, opened 2026-06-22
fix(go): replace WASM instances on state update to prevent memory growth

nicklasl:
## Problem

The Go provider's `PooledResolver` experiences unbounded WASM memory growth in production. Memory increases monotonically over time and is never reclaimed.

## Root cause

This is a heap fragmentation problem caused by the interaction between two operations that are individually stable:

1. **`SetResolverState`** — Allocates large contiguous blocks proportional to the resolver state size for the incoming protobuf payload, the decoded `ResolverState`, and its internal structures. During the transition, both old and new state are alive simultaneously.

2. **`ResolveProcess`** — Scatters small allocations across the WASM heap between state updates:
   - `FlagAssigned` entries pushed to the `AssignLogger`'s `crossbeam::SegQueue` (one per resolve with `apply=true`, containing cloned flag name, variant, resolve_id strings)
   - Resolve-info entries inserted into the `ResolveLogger`'s nested `papaya::HashMap`s (keyed by flag name, rule, variant, client credential, with atomic counters and schema samples)

**The fragmentation cycle:**

```
1. SetResolverState → allocates large block A for new state
2. ResolveProcess × N → small allocations (log entries) scatter through heap after A
3. SetResolverState → frees A, but freed region is peppered with live small allocations
4. dlmalloc cannot coalesce freed space → memory.grow for new state block B
5. WASM pages never shrink (spec limitation) → pages from A permanently retained
6. Repeat: each cycle the small allocations fragment the freed region, forcing new growth
```

Neither operation leaks in isolation — the Rust guest properly frees all allocations (Arc-based state replacement, SegQueue drain on flush, HashMap swap on checkpoint). The problem is that WASM linear memory pages, once grown, can never be returned to the host. The only way to reclaim them is to close the entire WASM module instance.

## Fix

`PooledResolver.SetResolverState` now replaces each slot's WASM instance with a fresh one instead of updating state in place:

1. Create a fresh resolver instance (via the existing supplier)
2. Set state on the fresh instance (clean heap, no fragmentation)
3. Write-lock the slot and swap in the fresh instance
4. Close the old instance (flushes pending logs, reclaims all WASM memory)

The write-lock swap ensures in-flight resolves complete on the old instance before it is retired. The fresh instance starts serving resolves immediately after the swap.

## Test plan

- [x] `TestWasmMemoryStableOnRepeatedSetResolverState` — state-only updates through PooledResolver (passed before and after fix)
- [x] `TestWasmMemoryStableOnRepeatedSetResolverStateWithResolves` — **new test** reproducing the production pattern: 200 state updates interleaved with 100 resolves each, through a PooledResolver. Grew +13.9MB before fix, 0 bytes growth after.
- [x] `TestWasmMemoryStableOnRepeatedResolveCalls` — existing resolve-only test (no regression)
- [x] All existing unit tests pass
- [x] Confirmed the same fragmentation pattern reproduces in the JS provider (same WASM binary)

### DIFF
--- a/openfeature-provider/go/confidence/internal/local_resolver/pool.go
+++ b/openfeature-provider/go/confidence/internal/local_resolver/pool.go
@@ -102,10 +102,37 @@ func (s *PooledResolver) ApplyFlags(request *resolver.ApplyFlagsRequest) error {
 }
 
 // SetResolverState implements LocalResolver.
+// WASM linear memory can grow but never shrink (spec limitation). Repeated
+// state updates on a long-lived instance cause heap fragmentation from
+// interleaved resolve allocations, leading to unbounded memory.grow calls.
+// Replacing instances reclaims the old linear memory entirely.
 func (s *PooledResolver) SetResolverState(request *wasm.SetResolverStateRequest) error {
-	return s.maintenance(func(lr LocalResolver) error {
-		return lr.SetResolverState(request)
-	})
+	s.mmu.Lock()
+	defer s.mmu.Unlock()
+
+	var errs []error
+	for i := range s.slots {
+		slot := &s.slots[i]
+
+		fresh := s.supplier()
+		if err := fresh.SetResolverState(request); err != nil {
+			fresh.Close(context.Background())
+			errs = append(errs, fmt.Errorf("slot %d: %w", i, err))
+			continue
+		}
+
+		slot.rw.Lock()
+		old := slot.lr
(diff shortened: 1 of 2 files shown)


nicklasl:
Closing in favor of a tracking issue — the instance-replacement approach is too intrusive across providers. The root cause is in the WASM guest's allocator behavior and deserves a broader design discussion.

PR #456, opened 2026-06-23
feat(wasm): add encrypted CDN state decryption support

nicklasl:
## Summary
- New proto message SetEncryptedResolverStateRequest
- New WASM guest export wasm_msg_guest_set_encrypted_resolver_state (AES-256-GCM, Tink NO_PREFIX)
- aes-gcm crate (~42K WASM binary increase)
- Encrypted test fixture for e2e tests
- Go WASM binary synced

Foundation PR - all provider PRs depend on this.


openfeature-provider/proto/confidence/wasm/messages.proto
```diff
@@ -17,6 +17,12 @@ message SetResolverStateRequest {
     confidence.flags.resolver.v1.Sdk sdk = 3;
 }
 
+message SetEncryptedResolverStateRequest {
```
andreas-karlsson, line 20: Is the account id in the state now?

wasm/rust-guest/src/lib.rs
```diff
@@ -140,6 +140,39 @@ wasm_msg_guest! {
         Ok(VOID)
     }
 
+    fn set_encrypted_resolver_state(request: SetEncryptedResolverStateRequest) -> WasmResult<Void> {
```
andreas-karlsson, line 143: Should we have a test for this?

openfeature-provider/proto/confidence/wasm/messages.proto
nicklasl: Well, it is part of the encrypted blob, yes.
Do you think we need to do some renaming?

wasm/rust-guest/src/lib.rs
nicklasl: I can try to set something up. not sure what kind of tests we have on the wasm level so far.
There are for sure e2e/integration tests in the providers.

openfeature-provider/proto/confidence/wasm/messages.proto
andreas-karlsson: No. Not really. Just wondered where it went.

andreas-karlsson (approved):

PR #457, opened 2026-06-23
feat(js): support encrypted CDN resolver state

nicklasl:
Depends on #456. Adds encryptionKey option, header+fallback detection, encrypted resolver state methods, crash recovery, and e2e test.


openfeature-provider/js/src/ConfidenceServerProviderLocal.e2e.test.ts
```diff
@@ -96,6 +96,48 @@ describe('ConfidenceServerProvider E2E tests', () => {
   });
 });
 
+describe('ConfidenceServerProvider Encrypted E2E tests', () => {
```
andreas-karlsson: I think this test would be conceptually more powerful if we just did a `describe.each` on the e2e test above. I.e. we skip this new test, but run the above description twice with the only modification of including the encryption key in one of the runs.

andreas-karlsson (approved):
Looks good! Approved with a small comment on the test.

PR #458, opened 2026-06-23
feat(java): support encrypted CDN resolver state

nicklasl:
Depends on #456. Adds encryptionKey config, header+fallback detection, setEncryptedResolverState through all resolver layers, and e2e test.


openfeature-provider/java/src/main/java/com/spotify/confidence/sdk/FlagsAdminStateFetcher.java
```diff
@@ -33,19 +34,34 @@ class FlagsAdminStateFetcher implements AccountStateProvider {
           com.spotify.confidence.sdk.flags.admin.v1.ResolverState.newBuilder()
               .build()
               .toByteArray());
+  private final AtomicReference<byte[]> rawCdnBytesHolder = new AtomicReference<>();
   private String accountId = "";
 
   public FlagsAdminStateFetcher(String clientSecret, HttpClientFactory httpClientFactory) {
```
nicklasl: I think we should remove this constructor and migrate the usages to send null

openfeature-provider/java/src/main/java/com/spotify/confidence/sdk/FlagsAdminStateFetcher.java
```diff
@@ -33,19 +34,34 @@ class FlagsAdminStateFetcher implements AccountStateProvider {
           com.spotify.confidence.sdk.flags.admin.v1.ResolverState.newBuilder()
               .build()
               .toByteArray());
+  private final AtomicReference<byte[]> rawCdnBytesHolder = new AtomicReference<>();
```
nicklasl: can we have a single holder and use that?

openfeature-provider/java/src/main/java/com/spotify/confidence/sdk/FlagsAdminStateFetcher.java
```diff
@@ -63,33 +57,37 @@ public void reload() {
...
         final byte[] bytes = stream.readAllBytes();
 
-        // Parse SetResolverStateRequest from CDN response
-        final var stateRequest =
-            com.spotify.confidence.sdk.wasm.Messages.SetResolverStateRequest.parseFrom(bytes);
-        this.accountId = stateRequest.getAccountId();
-
-        // Store the state bytes (already in bytes format)
-        rawResolverStateHolder.set(stateRequest.getState().toByteArray());
+        if (encryptionKey != null) {
```
nicklasl, line 83: should use `isEncrypted()`

openfeature-provider/java/src/main/java/com/spotify/confidence/sdk/RecoveringResolver.java
```diff
@@ -24,10 +24,13 @@ class RecoveringResolver implements LocalResolver {
 
   private record StateRecord(byte[] state, String accountId, Sdk sdk) {}
 
+  private record EncryptedStateRecord(byte[] encryptedState, byte[] encryptionKey, Sdk sdk) {}
```
nicklasl: This class doesn't need to support both encrypted and unencrypted at the same time (swapping between them).
If it is easier we can use the same backing record. If not, the "stateRecord" approach is scheduled to be removed later on.

openfeature-provider/java/src/main/java/com/spotify/confidence/sdk/FlagsAdminStateFetcher.java
```diff
@@ -63,33 +57,37 @@ public void reload() {
     }
   }
 
+  boolean isEncrypted() {
+    return isEncrypted();
```
nicklasl: seriously??

nicklasl (commented):
LGTM

andreas-karlsson (approved):

PR #459, opened 2026-06-23
feat(go): support encrypted CDN resolver state

nicklasl:
Depends on #456. Adds EncryptionKey config, header+fallback detection, SetEncryptedResolverState through all resolver layers, and e2e test.


openfeature-provider/go/confidence/internal/local_resolver/recover.go
```diff
@@ -39,7 +39,9 @@ type RecoveringResolver struct {
 	current atomic.Value // holds LocalResolver
 	broken  atomic.Bool  // indicates an instance has panicked
 
-	lastState atomic.Value // holds *wasm.SetResolverStateRequest
+	lastState          atomic.Value // holds *wasm.SetResolverStateRequest
+	lastEncryptedState atomic.Value // holds *wasm.SetEncryptedResolverStateRequest
+	useEncrypted       atomic.Bool
```
nicklasl: This class doesn't need to support both encrypted and unencrypted at the same time (swapping between them).
If it is easier we can use the same backing record. If not, the "stateRecord" approach is scheduled to be removed later on.

andreas-karlsson (approved):

nicklasl (commented):
LGTM

PR #460, opened 2026-06-23
feat(python): support encrypted CDN resolver state

nicklasl:
Depends on #456. Adds encryption_key parameter, header+fallback detection, set_encrypted_resolver_state through resolver layers, and e2e test.


andreas-karlsson (approved):

openfeature-provider/python/src/confidence/provider.py
```diff
@@ -217,12 +217,6 @@ def initialize(self, evaluation_context: EvaluationContext) -> None:
         Raises:
             Exception: If initialization fails.
         """
-        if not self._encryption_key:
-            logger.warning(
-                "No encryption_key provided. Falling back to unencrypted state. "
-                "An encryption key will be required in an upcoming version."
-            )
```
nicklasl: We should keep this warning log.

nicklasl:
LGTM

PR #464, opened 2026-06-24
feat(cloudflare): add sticky assignment support via KV store

vahidlazio:
## Summary
- Add sticky assignment (materialization) support to the Cloudflare resolver using Cloudflare KV
- Each assignment stored as a separate KV entry keyed by `mat:{unit}:{materialization}:{rule}` — no grouping, no merge, no read-modify-write races
- Components containing `:` are percent-encoded to prevent key parsing ambiguity
- Deployer creates KV namespace when `ENABLE_STICKY_ASSIGNMENTS` is set
- Optional TTL via `MATERIALIZATION_TTL_SECONDS`

## Key design decisions
- **KV-only** — KV's read-from-edge pattern is ideal for sticky assignments (read-heavy, write-once per user). Durable Objects don't add value with the `deferred_materializations` suspend/resume API.
- **Per-rule keys** — each rule assignment is a separate KV entry with a plain variant string as the value. Eliminates the need for grouping, merging, and read-modify-write cycles. All keys can be read/written independently.

## Test plan
- [x] Unit tests for key encoding and escaping (9 tests)
- [x] Clippy clean
- [x] End-to-end test with live A/B test flag on Cloudflare (confidence-test-wfs account)
- [x] Verified sticky consistency (same user → same variant across requests)
- [x] Verified KV entries use per-rule key format with plain variant values


confidence-cloudflare-resolver/src/materialization.rs
```diff
@@ -0,0 +1,618 @@
...
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+pub struct WriteRecord {
+    pub rule: String,
+    pub variant: String,
+}
+
+fn kv_key(unit: &str, materialization: &str) -> String {
+    format!("mat:{unit}:{materialization}")
```
andreas-karlsson: There is the potential problem of `:` in unit or materialization. I guess we control materialization, but not the unit. Should we escape it?

Also, have we considered using `"mat:{unit}:{materialization}:{rule}"` as the key? That would let us get rid of the read-modify-write race conditions, the merge and grouping. And all keys can be read/written in parallel so shouldn't hurt performance.

confidence-cloudflare-resolver/src/materialization.rs
```diff
@@ -0,0 +1,618 @@
...
+/// Each unique (unit, materialization) pair maps to a separate DO instance.
+pub async fn read_materializations_do(
+    ns: &ObjectNamespace,
+    records: &[MaterializationRecord],
+) -> Vec<MaterializationRecord> {
+    let groups = group_by_key(records);
+    let mut fetched: HashMap<Key, MaterializationData> = HashMap::new();
+
+    for (unit, materialization) in groups.keys() {
+        let id_name = format!("mat:{unit}:{materialization}");
```
andreas-karlsson: I was in general envisioning another DO pattern. The DO name should be based on only unit (or rather a combinations of selector and unit since different selectors might have the same unit value) this would mean that we have a DO instance per "user". If we then did the resolve inside that DO, in the resolve code we'd have access to a consistent synchronous(!) DB where we could read/write materializations as they are needed. The trade-off is of course that we'd need to restructure the resolve logic. It wouldn't use the same deferred_materializations mode etc.

confidence-cloudflare-resolver/src/lib.rs
```diff
@@ -168,6 +206,12 @@ pub async fn main(req: Request, env: Env, ctx: Context) -> Result<Response> {
         return Response::ok("")?.with_cors_headers(&allowed_origin_env);
     }
 
+    let mat_backend_for_writes = materialization::Backend::from_env(&env);
+    let mat_ttl: Option<u64> = env
+        .var("MATERIALIZATION_TTL_SECONDS")
```
andreas-karlsson, line 207: This only works for KV right? How should we handle cleanup in DO? We could leverage the DO Alarms API to schedule cleanup, but maybe messy.

andreas-karlsson (commented):
It's a nice integration in that it cleanly abstracts both storage APIs behind a common interface, but that also has the tradeoff that we're not using either API at it's full strength.

For DOs we could have one instance per user and if doing the resolve inside it we'd have access to a local synchronous(!) SQL DB. I.e. we could read/write materializations directly in the resolve logic without even messing with `await`.

For KV we could do an integration much more similar to the Java MaterializationHelper, i.e. when the first read is hit, await reading all materializations in parallel from different keys and then continue.

If we want to stick to the deferred_materializations API, I think the KV integration is the more sensible one and DO doesn't provide any extra benefits.

vahidlazio (commented):
Agreed — dropped the DO backend entirely. KV-only now, with per-rule keys.

confidence-cloudflare-resolver/src/materialization.rs
vahidlazio: Good catches, addressed both. Keys are now per-rule with colon percent-encoding. No more grouping, merging, or read-modify-write races.

confidence-cloudflare-resolver/src/materialization.rs
vahidlazio: Agreed — that would be the right DO architecture. Since it requires restructuring the resolve logic beyond deferred_materializations, dropped DO from this PR. KV is the right fit for the current API.

confidence-cloudflare-resolver/src/lib.rs
vahidlazio: Moot now that DO is removed. KV TTL via MATERIALIZATION_TTL_SECONDS works as expected.

confidence-cloudflare-resolver/src/materialization.rs
```diff
@@ -1,127 +1,43 @@
...
 
-    for ((unit, materialization), group_records) in &groups {
-        let Some(data) = fetched.get(&(unit.clone(), materialization.clone())) else {
-            continue;
-        };
-
-        for record in group_records {
+    for record in records {
+        let key = kv_key(record);
+        if let Ok(Some(value)) = kv.get(&key).text().await {
```
andreas-karlsson: It looks like we're serializing all reads? Consider using `futures::future::join_all`

confidence-cloudflare-resolver/src/materialization.rs
```diff
@@ -0,0 +1,140 @@
...
+            record.variant.clone()
+        };
+
+        let builder = kv.put(&key, value);
+        let builder = match (builder, ttl_seconds) {
+            (Ok(b), Some(ttl)) => Ok(b.expiration_ttl(ttl)),
+            (builder, _) => builder,
+        };
+        if let Ok(b) = builder {
+            let _ = b.execute().await;
```
andreas-karlsson: Here we also serialise writes, maybe not as bad as it happens in wait_until though.

confidence-cloudflare-resolver/src/materialization.rs
```diff
@@ -0,0 +1,140 @@
...
+
+fn encode_component(s: &str) -> String {
+    s.replace('%', "%25").replace(':', "%3A")
+}
+
+fn kv_key(record: &MaterializationRecord) -> String {
+    let unit = encode_component(&record.unit);
+    let mat = encode_component(&record.materialization);
+    if record.rule.is_empty() {
+        format!("mat:{unit}:{mat}:__included__")
```
andreas-karlsson: Are we ever writing these sentinel values? I don't think so. Either you would have to take care if it yourself, i.e. write the sentinel on each write, but I think the better option is to handle empty rule by doing a prefix search. I think the prefix search has a limit clause that we can set to one so that it doesn't fetch a bunch of  records, but it might still do that underneath. Would need some digging in CF docs or so.

confidence-cloudflare-resolver/deployer/script.sh
```diff
@@ -423,6 +424,57 @@ fi
...
+    if [ -n "$WORKER_NAME_PREFIX" ]; then
+        MAT_KV_TITLE="${WORKER_NAME_PREFIX}-resolver-materializations"
+    else
+        MAT_KV_TITLE="resolver-materializations"
+    fi
+
+    echo "🔍 Checking if KV namespace '$MAT_KV_TITLE' exists..."
+    MAT_KV_LIST=$(curl -sS -w "%{http_code}" \
+        -H "Authorization: Bearer ${CLOUDFLARE_API_TOKEN}" \
+        "https://api.cloudflare.com/client/v4/accounts/${CLOUDFLARE_ACCOUNT_ID}/storage/kv/namespaces?per_page=100")
```
andreas-karlsson: Nit. But could a customer have more than 100 namespaces? If so, what would be the consequences? We create a new ns on each deploy? Tricky bug if it ever happened. Maybe a safe option is to verify we don't get a full 100 results and otherwise error? (Or impl. proper page search)

andreas-karlsson (approved):

PR #483, opened 2026-07-14
feat: multi-destination flag log routing

vahidlazio:
## Summary

- Add `LogDestination` enum (`SPOTIFY_EDGE` / `CLOUDFLARE`) to `ClientResolverState` proto (repeated field, backwards compatible — empty defaults to Edge)
- Parse `log_destinations` in Rust `ResolverState::from_proto` and expose as typed enum
- Route flag logs in Cloudflare worker queue consumer based on account's configured destinations
- Order-dependent routing: first destination is primary, second is fallback on HTTP 4xx/5xx or network error
- Wire compatibility test proving tag 4 on `ClientResolverState` is safely skipped by existing WASM parsers

### Deployment order
1. Merge this PR (backwards compat — all accounts default to Edge)
2. Set `log_destinations=[CLOUDFLARE]` on test account via flags-admin (separate PR)
3. Verify logs land in R2, then gradual rollout

### Flags-admin changes (separate PR)
- Add `log_destinations` to admin's copy of `resolver.proto`
- Populate field when building `ClientResolverState` for CDN push
- Account-level setting API

## Test plan
- [x] All existing unit tests pass
- [x] All crates compile (`confidence_resolver`, `confidence-cloudflare-resolver`)
- [x] Wire compatibility test for proto tag safety
- [ ] Build test state proto with `log_destination=CLOUDFLARE`, deploy CF resolver, verify logs routed to ingest worker


confidence-cloudflare-resolver/src/lib.rs
```diff
@@ -521,19 +612,61 @@ async fn update_prometheus_kv(kv: &kv::KvStore, req: &WriteFlagLogsRequest) {
...
+        let mut body = Vec::new();
+        // field 1: string account_id, tag = 0x0a
+        body.push(0x0a);
+        encode_varint(account_bytes.len() as u64, &mut body);
+        body.extend_from_slice(account_bytes);
+        // field 2: WriteFlagLogsRequest batch, tag = 0x12
+        body.push(0x12);
+        encode_varint(batch_buf.len() as u64, &mut body);
+        body.extend_from_slice(&batch_buf);
+
```
fabriziodemaria, line 695: I am sure we had a good reason not to have proto here and using proto-native builders and so on? This hand-rolled payload is hard to read

confidence-cloudflare-resolver/src/lib.rs
```diff
@@ -486,7 +551,33 @@ pub async fn consume_flag_logs_queue(
...
+        let result = send_flags_logs(client_secret, &req, primary_url, primary_acct).await;
+
+        if let Some(fb) = fallback {
+            if result.is_err() || result.as_ref().is_ok_and(|r| r.status_code() >= 400) {
+                let fb_url = log_destination_url(&fb);
+                let fb_acct = match fb {
+                    LogDestination::Edge => None,
+                    _ => Some(account_id),
+                };
+                let _ = send_flags_logs(client_secret, &req, fb_url, fb_acct).await;
```
fabriziodemaria: If both primary and fallback fail, logs are permanently lost with zero observability. Adding a `console_log!` on fallback attempt and on fallback failure would help with production debugging:

```rust
console_log!("primary log delivery failed, trying fallback");
match send_flags_logs(client_secret, &req, fb_url, fb_acct).await {
    Err(e) => console_log!("fallback log delivery also failed: {:?}", e),
    Ok(r) if r.status_code() >= 400 => console_log!("fallback returned status {}", r.status_code()),
    _ => {}
}
```

confidence-cloudflare-resolver/src/lib.rs
```diff
@@ -486,7 +551,33 @@ pub async fn consume_flag_logs_queue(
...
+
+        let primary_url = log_destination_url(&primary);
+        let primary_acct = match primary {
+            LogDestination::Edge => None,
+            _ => Some(account_id),
+        };
+        let result = send_flags_logs(client_secret, &req, primary_url, primary_acct).await;
+
+        if let Some(fb) = fallback {
+            if result.is_err() || result.as_ref().is_ok_and(|r| r.status_code() >= 400) {
```
fabriziodemaria: **[suggestion]** The `>= 400` threshold means client errors (400, 403, etc.) also trigger fallback. A 400 Bad Request likely means the request body itself is malformed, which will probably fail identically on the fallback. Consider whether `>= 500` (server errors only) is more appropriate, or whether this is intentional (e.g. the Cloudflare ingestor might return 4xx for transient auth issues that the Edge endpoint handles differently).

confidence-resolver/protos/confidence/flags/admin/v1/resolver.proto
```diff
@@ -87,6 +87,12 @@ message ResolverState {
 
   repeated BloomFilter bloom_filters = 10;
 
+  enum LogDestination {
+    LOG_DESTINATION_UNSPECIFIED = 0;
+    LOG_DESTINATION_SPOTIFY_EDGE = 1;
+    LOG_DESTINATION_CLOUDFLARE = 2;
+  }
```
fabriziodemaria, line 99: **[minor]** The enum is defined here but no `repeated LogDestination log_destinations = N;` field is added to `ResolverState`. The field only exists in the manually-defined `ClientResolverState` prost struct in the Cloudflare resolver (at tag 4). Someone reading this proto won't see how destinations are actually transmitted.

Consider adding a comment to the enum noting where the field lives, or adding a `reserved 11;` + comment to `ResolverState` if you plan to eventually formalize it.

confidence-cloudflare-resolver/src/lib.rs
```diff
@@ -486,7 +551,33 @@ pub async fn consume_flag_logs_queue(
...
+        } else {
+            (destinations[0], None)
+        };
+
+        let primary_url = log_destination_url(&primary);
+        let primary_acct = match primary {
+            LogDestination::Edge => None,
+            _ => Some(account_id),
+        };
+        let result = send_flags_logs(client_secret, &req, primary_url, primary_acct).await;
```
fabriziodemaria: The old code ended with `send_flags_logs(...).await?;` — a transport error propagated out of `consume_flag_logs_queue`, so Cloudflare Queues would retry the batch. The new code captures `result`, optionally tries the fallback, and then unconditionally returns `Ok(())`, so the batch is acked even when every delivery attempt failed. Combined with the silent fallback below, a delivery outage now permanently drops logs with no retry and no trace. 

Not sure how to combine our retry with the CloudFlare Queue retries

Cargo.lock
```diff
@@ -183,15 +157,15 @@ checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
...
-version = "2.9.3"
+version = "2.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "34efbcccd345379ca2868b2b2c9d3782e9cc58ba87bc7d79d5b53d9c9ae6f25d"
+checksum = "b588b76d00fde79687d7646a9b5bdf3cc0f655e0bbd080335a95d7e96f3587da"
 
 [[package]]
 name = "bitvec"
-version = "1.0.1"
+version = "1.1.1"
```
fabriziodemaria: **[major]** This PR embeds a wholesale `cargo update`: 172 packages bumped, 17 added, 13 removed — with no `Cargo.toml` changes at all. Since the manifests are unchanged relative to main, main's lockfile still resolves; none of these bumps are required by the feature.

Notable bumps riding along: `worker` 0.6.1→0.6.7 (the Workers runtime this component runs on), `bitvec` 1.0.1→1.1.1 (resolver-core bitset evaluation, baked into the committed Go WASM binary), `wasm-bindgen` 0.2.100→0.2.127, `aws-lc-sys` 0.39.1→0.44.0. So the shipped artifacts change far more than the source diff suggests.

Suggestion: revert to main's lockfile (`git checkout origin/main -- Cargo.lock`), re-run `make sync-wasm-go` so the committed WASM matches, and land the dependency refresh as its own PR where it can be reviewed and bisected independently. If the bumps are intentionally part of this change, please say so in the PR description.

fabriziodemaria (commented):
My only really blocking comment is to make sure we have EDGE as the default: I noticed now there is CF if the state has no "destination" in it

confidence-cloudflare-resolver/src/lib.rs
```diff
@@ -574,19 +612,61 @@ async fn update_prometheus_kv(kv: &kv::KvStore, req: &WriteFlagLogsRequest) {
...
-    init.with_headers(headers);
     init.with_method(Method::Post);
-    let json = serde_json::to_string(&message)?;
-    init.with_body(Some(json.into()));
-    let request = Request::new_with_init(resolve_url, &init)?;
-    let response = Fetch::Request(request).send().await;
-    response
+
+    if let Some(account) = account_id {
+        // Cloudflare ingestor: encode as IngestFlagLogsRequest protobuf
```
fabriziodemaria, line 681: `IngestFlagLogsRequest, WriteFlagLogsRequest`

What's the difference between these two?

vahidlazio:
@fabriziodemaria hmm default is spotify edge i believe.

fabriziodemaria:
>  hmm default is spotify edge i believe.

Right, I got confused because in my mind Cloudflare is "Edge". So perhaps we want to map the enum to SpotifyEdge rather than just Edge.

fabriziodemaria (approved):

PR #495, opened 2026-07-31
fix(resolver): reject empty range rules

### LINKED ISSUE #296: Rust resolver treats empty range rule as always matching
## Problem

An empty range rule (`rangeRule {}` with no start or end bounds) matches any value in the Rust resolver. Java treats it as non-matching.

## Rust behavior

`confidence-resolver/src/value.rs` ~L162-187:

```rust
fn evaluate_range_rule(
    range_rule: &targeting::RangeRule,
    context_value: &targeting::Value,
) -> bool {
    let after_start = match &range_rule.start {
        Some(...) => ...,
        _ => false,
    };
    let before_end = match &range_rule.end {
        Some(...) => ...,
        _ => false,
    };
    (range_rule.start.is_none() || after_start) && (range_rule.end.is_none() || before_end)
}
```

When both `start` and `end` are `None`:
- `range_rule.start.is_none()` → `true`
- `range_rule.end.is_none()` → `true`
- Result: `true && true` = `true` (matches)

## Java behavior

`EvalUtil.java` ~L167-178:

```java
public static boolean isInRange(Targeting.RangeRule rangeRule, Targeting.Value value) {
    if (value == null) {
        return false;
    }
    try {
        return Util.hasOverlap(value, rangeRule);
    } catch (IllegalArgumentException e) {
        return false;
    }
}
```

Java delegates to `Util.hasOverlap()` which requires at least one bound to be set. An empty range has `VALUE_NOT_SET` for its expected type:

```java
private static ValueCase getExpectedPrimitiveType(Targeting.RangeRule rangeRule) {
    if (rangeRule.hasStartInclusive()) return rangeRule.getStartInclusive().getValueCase();
    else if (rangeRule.hasStartExclusive()) return rangeRule.getStartExclusive().getValueCase();
    else if (rangeRule.hasEndInclusive()) return rangeRule.getEndInclusive().getValueCase();
    else if (rangeRule.hasEndExclusive()) return rangeRule.getEndExclusive().getValueCase();
    else return ValueCase.VALUE_NOT_SET;
}
```

With no bounds, the type check fails and the range does not match → `NO_SEGMENT_MATCH`.

## Affected spec tests

- `empty_range_rule`

Pybsama:
## Why

The Rust resolver currently treats a range rule with neither a start nor an end
bound as matching every context value. That differs from the shared resolver
specification and the Java resolver, where an empty range is a normal
non-match.

The final boolean expression considered each absent bound automatically
satisfied, so an entirely empty rule evaluated to `true`.

## What

- Return `false` when both range bounds are absent, while leaving all valid
  one-sided, inclusive, and exclusive comparisons unchanged.
- Remove the Rust-only expected-result override from the existing
  `empty_range_rule` resolver spec so Rust now uses the shared
  `NO_SEGMENT_MATCH` expectation.
- Rebuild the Go provider's committed embedded WASM resolver.

## Validation

- TDD regression: removing the Rust override failed with
  `expected RESOLVE_REASON_NO_SEGMENT_MATCH, got MATCH` before the production
  change, then passed after the guard was added.
- Range-focused Rust tests: 21 passed.
- Core resolver release suite on the repository-pinned Rust 1.90.0 toolchain:
  371 passed.
- Core `rustfmt` and release Clippy on Rust 1.90.0.
- WASM guest `rustfmt` and release Clippy for `wasm32-unknown-unknown` on
  Rust 1.90.0.
- The rebuilt WASM asset byte-for-byte matches the pinned-toolchain build
  output (SHA-256
  `2a51a26f8daba8d9ea5d66977ddec9a87b5b9ec049be6f9e13df16b02f7c945f`).
- Go packages that consume and exercise the embedded resolver:
  `confidence/internal/local_resolver` and
  `confidence/internal/flag_logger` both passed.

The repository-wide Go target also contains real-backend E2E cases. In an
unconfigured local environment those existing cases stop at a missing
`ClientSecret` or mock-support-server 404, so they are not reported as passing
here.

## Compatibility

No protobuf, dependency, or public API changes. Empty ranges now match the
shared cross-language contract; ranges with at least one bound keep their
existing behavior.

Closes #296.


Pybsama:
Thanks @nicklasl for picking this up in #543 and preserving the original authorship and cherry-pick provenance.

I verified that the two source/spec changes in `c09c11e` have the same stable patch ID as my original commit, that #543 merges cleanly with current `main`, and that its Build & Test plus CodeQL checks are all passing. The Build & Test failure shown on this fork PR is from the fork workflow receiving no backend secrets (`ClientSecret is required` / mock endpoint 404); the affected local resolver package itself passed.

Since #543 carries the current-main WASM artifact and is now the upstream review vehicle, I am closing this Draft to keep review focused there. This PR remains available as the original implementation and validation context, and I will follow #543.

PR #501, opened 2026-08-03
feat(cloudflare): add FORCE_APPLY option to control apply behavior

vahidlazio:
## Summary

Adds a `FORCE_APPLY` configuration option to the Cloudflare Worker resolver, defaulting to `true`.

**`FORCE_APPLY=true` (default):**
- Every `flags:resolve` is treated as `apply=true`, so assignments are logged at resolve time.
- The resolve token is stripped from the response. SDKs that resolved with `apply=false` then send their background `flags:apply` with an empty token, which hits the existing no-op path — no double logging, regardless of backend dedupe behavior.

**`FORCE_APPLY=false` (opt-out via deploy script):**
- The `apply` value sent by the SDK is respected and the resolve token is returned, keeping the deferred-apply flow (#496) fully functional.

## Deployer

- New optional `FORCE_APPLY` env var, validated to be `true`/`false` (case-normalized), injected into the `[vars]` table of `wrangler.toml` following the existing `ALLOWED_ORIGIN` pattern (idempotent on re-deploys).
- When unset, nothing is written and the worker's built-in default (`true`) applies.

## Notes

- Transitional edge: resolve tokens issued before a deployment enables `FORCE_APPLY` can still be applied via `flags:apply`; only affects the window around a config change.
- Verified: wasm target compiles, `make lint` (clippy `-D warnings`) passes, deployer TOML injection exercised with the actual script fragment.


confidence-cloudflare-resolver/src/lib.rs
```diff
@@ -323,12 +335,18 @@ pub async fn main(req: Request, env: Env, ctx: Context) -> Result<Response> {
...
                                             MAT_WRITES.with(|f| {
                                                 *f.borrow_mut() = Some(writes);
                                             });
                                         }
+                                        // Flags are already applied; an empty
+                                        // token makes SDK background applies
+                                        // hit the no-op path in flags:apply.
+                                        if force_apply {
+                                            response.resolve_token = Default::default();
+                                        }
```
andreas-karlsson: I was a bit confused about this initially. With apply true on the request this isn't necessary right?

andreas-karlsson: I think it'd be a bit nicer to omit it.

andreas-karlsson (approved):

PR #512, opened 2026-08-11
fix(openfeature-provider/go): prevent silent SIGSEGV when recovery closes an instance mid-resolve

### LINKED ISSUE #511: Silent process crash (SIGSEGV, no traceback): recovery closes the wasm instance while concurrent resolves still use it
**Version:** openfeature-provider/go v0.20.1 (latest), wazero v1.12.0, go1.26.2, linux/arm64

## Symptom

Our Go service dies with exit code 139 and no Go traceback, roughly 16 times per week across a ~150-pod fleet with heavy flag-evaluation volume. The only evidence is a kernel core dump. We captured and analyzed cores from two independent crashes on different builds.

## Core dump analysis

Both cores show the same synchronous nil-pointer dereference. The faulting instruction is the itab method load of a nil `api.Function` interface (`ldr xN, [xN, #24]` with `xN = 0`, fault address `0x18`):

- Crash A: pc = `local_resolver.allocAndWrite+76` (wasm.go:264, `allocFn.Call(...)`). Interrupted stack: `RecoveringResolver.ResolveProcess.func1` → `WasmResolver.ResolveProcess` → `call` → `transfer` → `allocAndWrite`.
- Crash B: pc = `(*WasmResolver).call+428` (`fn.Call(...)`). Same registers, same fault address.

In both cases the Go runtime could not attribute the fault and died via `sigfwdgo → dieFromSignal(SIGSEGV)`, so the process exits silently instead of panicking with a traceback.

## Mechanism

`PooledResolver` slots hold `RecoveringResolver(WasmResolver)` and each resolve takes only a shared `RLock` (pool.go), so two resolves can run on one slot concurrently. When one of them panics (a guest trap), `withRecover` fires `startRecreate` (recover.go), which closes the old wazero instance in a background goroutine while the concurrent resolve is still mid-call on it. `WasmResolver.call` holds `r.mu` for the duration of a call, but `Close` does not take `r.mu`, so the close lands mid-call.

A resolve racing with that close can observe a nil function handle. Notably, a fully closed instance is not enough to reproduce this: `ExportedFunction` on a fully closed module returns a non-nil handle whose `Call` fails with a recoverable error. Both cores show the nil itab load, so the nil arises in the close race window, and `exportedFunction` (wasm.go:52) would even cache the nil in `fnCache`. The next `fn.Call` dereferences the nil interface.

The initial guest panic is invisible to operators: `withRecover` converts it into an evaluation error, which OpenFeature swallows into default values. The follow-on nil deref then kills the whole process.

## Suggested fixes

1. Serialize `Close` with in-flight calls: take `r.mu` in `WasmResolver.Close` around `instance.Close`. This closes the race window.
2. Never cache a nil handle in `fnCache`, and have `exportedFunction` panic with a descriptive error instead of returning nil. `withRecover` already catches panics on that path, so this converts silent process death into a recovered error.

We can share the full register dumps and stack walks on request. The raw cores contain production data, so we cannot attach them.

tmlye:
Fixes #511.

## Problem

`RecoveringResolver.startRecreate` closes the old wazero instance in a background goroutine. A concurrent resolve can still be mid-call on that instance, because pool slots only take a shared `RLock` per resolve. `WasmResolver.call` holds `r.mu` for the duration of a call, but `Close` does not take it. The close therefore lands mid-call. A resolve that races with the close can get a nil `api.Function` from `ExportedFunction`, and `exportedFunction` caches the nil. The next `fn.Call` dereferences the nil interface. The Go runtime cannot attribute that fault, so it kills the process silently (`sigfwdgo → dieFromSignal(SIGSEGV)`, exit 139, no traceback). See #511 for the core dump analysis.

## Changes

1. `WasmResolver.Close` now takes `r.mu` around `instance.Close`. This serializes close with in-flight calls and removes the race window. Primary fix.
2. `exportedFunction` panics with a clear message instead of returning nil, and never caches a nil handle. This is hardening. `withRecover` already catches panics on this path, so a residual nil becomes a recovered evaluation error instead of silent process death.

## Testing

- New `TestResolveOnClosedInstancePanics` pins the closed-instance contract. A resolve on a closed instance must end in a recoverable panic. It must never reach `fn.Call` through a nil handle. The test closes the raw instance directly, because `WasmResolver.Close` flushes logs first and that would warm `fnCache`. The direct close exercises the post-close `ExportedFunction` path.
- `go test ./confidence/internal/local_resolver/` passes.
- Honest limitation: the silent-death mode only reproduces under the production race, where the close lands inside the mid-call window. A fully closed instance returns a non-nil handle, and its `Call` fails with a recoverable error. The test therefore pins the contract, not the exact race. The two production cores in #511 are the evidence for the race itself.
- The `TestFlagResolve_*` e2e tests fail with "ClientSecret is required" both with and without this change. They need credentials this environment does not have.


fabriziodemaria (approved):
Reviewed — this looks like the right fix for the cored crashes in #511.

**Minor follow-up (non-blocking)**: consider also skipping cached nil entries on load, not just guarding the fresh lookup path:

```go
if fn, ok := r.fnCache.Load(name); ok && fn != nil {
    return fn.(api.Function)
}
```

Prevents a stale cached nil from ever reaching `fn.Call` if one was stored before this fix.

PR #517, opened 2026-08-13
feat: flag apply event deduplication in core resolver and WASM guest

vahidlazio:
## Summary
- Adds `apply_dedup` module to `confidence-resolver`: a TTL-based in-memory dedup cache for flag apply events
- Integrates it in the WASM guest's `log_assign` path — an assignment identical to one already logged within the TTL window (2 min) is skipped
- Eliminates redundant apply events from repeated resolves (SPA page transitions, polling SDKs): in a 30 s load test the same-user workload went from **458 MB / 1.13 M events shipped to 3 KB / 3 events**, with 5.5x lower peak RSS and 7% less total CPU — at a throughput cost indistinguishable from noise (≤1.5%)

## Dedup semantics
There is exactly **one key schema**, defined in `compute_applied_flag_dedup_hash` over a borrowed `AppliedFlagRef` view (the `AssignedFlag` path delegates through `From`, so WASM and the future Cloudflare path cannot drift). The key is the **full apply identity — every field except `apply_time`**, including all `fallthrough_assignments` entries. No field may be dropped: `assignment_id` is an opaque user-chosen label (rules can share names like "control", and an assignment can be repointed to a new variant without renaming), and fallthrough-only / no-unit shapes carry per-user identity in the fallthrough entries. Dedup only runs when `apply = true`; empty apply envelopes bypass it and log exactly as before.

**Config:** TTL = 120 s, cap = 100,000 entries (~2 MB, grown lazily so small deployments stay cache-resident). At the cap new entries pass through unlogged-but-undeduped (fail-open: over-log, never drop).

**Experimental, opt-in (off by default)** via `enable_apply_dedup` on `SetResolverStateRequest`. With dedup off, `log_assign` is a single atomic load + passthrough — behavior and performance identical to pre-dedup (verified end-to-end). Provider options (all marked experimental):
- Go: `ProviderConfig.EnableApplyDedup` / `WithEnableApplyDedup()`
- JS: `ProviderOptions.enableApplyDedup`
- Java: `LocalProviderConfig.Builder.enableApplyDedup(...)`
- Python: `ConfidenceOpenFeatureProvider(enable_apply_dedup=...)`

The Rust provider (native, no WASM) and Ruby provider (remote resolver) are not affected by this PR.

## Architecture
- **Resolve path** pays only: hash (murmur-style word-at-a-time avalanche, zero alloc) + one `contains_key` probe + insert on miss. No TTL comparison, no timestamp writes, no scans.
- **Zero clones on the apply path**: `FlagToApply` borrows the assignment (`&AssignedFlag`); assignments are borrowed straight from resolved values / the resolve token, and strings are copied only when an event is actually logged into the proto.
- **Expiry** runs solely on the host-driven log flush cycle: one O(n) sweep, self-throttled to at most once per 10 s. Insert timestamps clamp to the last sweep time so deferred applies with old skew-adjusted times aren't inserted pre-expired.
- **Hash quality**: each field is length-prefixed (injective encoding — fixes a deterministic short-tail vs full-word collision that could suppress a crafted key's events) and every 8-byte word is avalanched (murmur3 fmix64 constants); regression tests cover the collision pair, boundary shifts, and 200K structured keys.
- **Keep-mask** is an inline `u64` for batches ≤ 64 flags (no allocation) and spills to a `Vec<u64>` beyond (resolver allows 1000 flags/request).

## Performance

Microbenchmarks (release, `--ignored`): dedup-hit path 177 ns vs 1,033 ns baseline (83% faster); worst-case all-unique overhead ~350 ns/resolve.

End-to-end Go provider load test (30 s, 1 thread, Docker, mock backend):

| Scenario | Metric | With dedup | Without dedup |
|---|---|---|---|
| Same user (all duplicates) | throughput | 32.8K ops/s | 33.0K ops/s (−0.8%, ≈ noise) |
| | apply events shipped | **3** | **1,131,774** |
| | bytes shipped | **3 KB** | **458 MB** |
| | peak RSS | **48 MB** | **267 MB** |
| | total CPU | **37.6 s** | **40.2 s** |
| All unique (worst case) | throughput | 42.5K ops/s | 43.2K ops/s (−1.5%, ≈ noise) |
| | apply events shipped | 1 per resolve | 1 per resolve |
| | peak RSS / CPU | 242 MB / 38.1 s | 232 MB / 38.2 s |

## Behavior changes
- Repeated resolves of the same flag+user within the TTL produce one apply event instead of one per resolve. Provider tests updated: same-user cases assert exact bounds (JS/Java single instance = 1; Go pooled = strictly fewer than resolves), new distinct-user cases assert dedup never collapses different users, shutdown-drain tests use per-task keys.
- Go embedded WASM binary re-synced (Docker-built to match CI).

## Bench tooling (included)
- `go-bench`: new `-unique-keys` mode (dedup worst case) and CPU/peak-RSS reporting; Dockerfile Go image bumped to 1.25 to match go.mod (was unbuildable).
- Rust perf/stress benchmarks are `#[ignore]` — run with `cargo test --release -- --ignored`.

## Follow-up (separate PR)
Cloudflare Worker KV-backed dedup, opt-in via `ENABLE_APPLY_DEDUP` (default off); deployer creates the KV namespace when opted in. `compute_applied_flag_dedup_hash` is already exported for it.

## Test plan
- [x] 27 dedup unit tests: TTL + sweep expiry, sweep throttling, per-field differentiation incl. fallthrough-only users and variant/rule changes under a shared assignment_id, capacity behavior, >64-flag mixed batches, empty-envelope passthrough, AssignLogger integration, hash determinism/consistency
- [x] Collision regression: 200K structured keys, zero collisions
- [x] clippy + fmt clean (`confidence_resolver`, `rust-guest` wasm32, `confidence-cloudflare-resolver` wasm32); full suite green in CI across all providers
- [x] End-to-end load tests, both scenarios, zero errors across ~9M resolves


confidence-resolver/src/apply_dedup.rs
```diff
@@ -0,0 +1,1156 @@
...
+}
+
+type IdentityBuildHasher = BuildHasherDefault<IdentityHasher>;
+
+/// Streaming word-at-a-time hasher with murmur3-style per-word avalanche.
+/// Strong mixing is required because hashes are used directly as map keys
+/// (identity hasher) and inputs are highly structured.
+struct DedupHasher(u64);
+
+impl DedupHasher {
```
nicklasl, line 36: Is there nothing built in we can use instead of defining this ourselves?

openfeature-provider/js/src/ConfidenceServerProviderLocal.ts
```diff
@@ -46,6 +46,12 @@ export interface ProviderOptions {
   flushInterval?: number;
   fetch?: typeof fetch;
   materializationStore?: MaterializationStore | 'CONFIDENCE_REMOTE_STORE';
+  /**
+   * Disable apply-event deduplication in the WASM resolver. Dedup is enabled
+   * by default: repeated identical assignments within a short TTL window are
+   * logged once.
+   */
+  disableApplyDedup?: boolean;
```
nicklasl: This goes for all the providers:
My take is that we should inverse this to
```suggestion
  enableApplyDeduplication?: boolean;
```
and also default it to be **opt-in** until we have run this on our own SDK's in production.

Also, suggesting marking it as experimental so we can change it without consider it a breaking change.

nicklasl (approved):

PR #524, opened 2026-08-19
feat: event tracking support

vahidlazio:
## Summary

- Adds a new WASM module (`confidence_event_engine.wasm`, 56KB) that batches OpenFeature `track()` events for server-side providers (JS, Java, Go, Python)
- Uses the same lock-free `SegQueue` + bounded flush pattern as the resolver's `AssignLogger` — proven, no new patterns to audit
- Providers call `track_event()` to queue events and `bounded_flush_events()` to drain up to 4MB batches, then POST to `events.confidence.dev/v1/events:publish` (matching Android SDK format)

### New components

| Component | Purpose |
|-----------|---------|
| `confidence-event-engine/` | Core Rust crate: `EventLogger` with lock-free batching, 8 unit tests |
| `wasm/event-guest/` | WASM guest binary (56KB, zero host imports, fully self-contained) |
| `openfeature-provider/proto/confidence/events/v1/` | Shared proto: `Event`, `PublishEventsRequest` |
| `tests/event-engine-e2e/{go,js,python}/` | E2E load tests + memory/perf benchmarks against real API |

### WASM exports

| Export | Input → Output | Purpose |
|--------|----------------|---------|
| `wasm_msg_guest_track_event` | `Event → Void` | Push event into batch queue |
| `wasm_msg_guest_bounded_flush_events` | `Void → PublishEventsRequest` | Drain up to 4MB, return network-ready batch |

### Provider integration pattern

```
Provider.track(eventName, context, details)
  → merge context into payload under "context" key (matching Android SDK)
  → call wasm_msg_guest_track_event(Event{...})

Provider flush timer (~15s)
  → call wasm_msg_guest_bounded_flush_events(Void)
  → decode PublishEventsRequest from protobuf
  → serialize to JSON: { clientSecret, sdk, sendTime, events }
  → POST to events.confidence.dev/v1/events:publish
```

### Performance (Apple M1 Max)

| Provider | Track throughput | Per-event latency |
|----------|-----------------|-------------------|
| Go (wazero) | 1.68M events/sec | 593ns |
| JS (Node WebAssembly) | 302K events/sec | 3.3µs |
| Python (wasmtime) | 16K events/sec | 62µs |

- Zero memory leaks confirmed across 50K+ event cycles
- 10K events flush in a single 490KB batch in <10ms

## Test plan

- [x] `cargo test -p confidence-event-engine` — 8 unit tests (size limits, cross-flush persistence, data integrity)
- [x] `cargo build -p event-guest --target wasm32-unknown-unknown --profile wasm` — WASM compiles (56KB)
- [x] `cargo clippy` + `cargo fmt` — clean on both crates
- [x] Go e2e: load WASM → track 1K events → flush → POST to real events API (HTTP 200)
- [x] JS e2e: same flow, zero external dependencies
- [x] Python e2e: same flow via wasmtime
- [x] Go memory tests: WASM linear memory stable, Go heap stable, no leaks
- [x] Go benchmarks: track_event 593ns/op, flush scales linearly
- [x] Existing resolver WASM still builds and works (no regressions)


tests/event-engine-e2e/go/main.go
```diff
@@ -0,0 +1,324 @@
...
+	"os"
+	"time"
+
+	"github.com/tetratelabs/wazero"
+	"github.com/tetratelabs/wazero/api"
+	"google.golang.org/protobuf/encoding/protowire"
+)
+
+const (
+	clientSecret = "DxAyUiTjjlfzLNnwjsD25c4uaHaCafp8"
```
nicklasl: committed clientSecret...
we should rotate this and if we mean to run the e2e tests automatically put them in the gh secrets.

Dockerfile
```diff
@@ -45,6 +45,8 @@ COPY confidence-resolver/Cargo.toml ./confidence-resolver/
 COPY confidence-cloudflare-resolver/Cargo.toml ./confidence-cloudflare-resolver/
 COPY wasm-msg/Cargo.toml ./wasm-msg/
 COPY wasm/rust-guest/Cargo.toml ./wasm/rust-guest/
+COPY wasm/event-guest/Cargo.toml ./wasm/event-guest/
```
nicklasl, line 48: Can we also make sure the event engine is included in the Docker stages that actually run from the final `all` target? I guess copying the crate into the build context is the first step, but since this PR adds a committed `confidence_event_engine.wasm`, I just want to make sure CI also builds/lints/tests the standalone module rather than only making its sources available.

wasm/confidence_event_engine.wasm
nicklasl, line 1: I don't think this should this be committed.

confidence-event-engine/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "confidence-event-engine"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[dependencies]
+prost = { version = "0.13", default-features = false }
+prost-types = { version = "0.13", default-features = false }
+crossbeam-queue = "0.3"
```
nicklasl: existing resolver pins `crossbeam-queue = { version = "0.3.12", default-features = false, features = ["alloc"] }`
is the discrepancy intentional?

wasm/event-guest/src/lib.rs
```diff
@@ -0,0 +1,22 @@
+use std::sync::LazyLock;
+
+use confidence_event_engine::event_logger::EventLogger;
+use confidence_event_engine::proto::confidence::events::v1::{Event, PublishEventsRequest, Void};
+use wasm_msg::wasm_msg_guest;
+use wasm_msg::WasmResult;
+
+const LOG_TARGET_BYTES: usize = 4 * 1024 * 1024; // 4 MB
```
nicklasl: let's put this a bit lower to give it some head room for other things (clientsecret, sdk info etc)

wasm/event-guest/src/lib.rs
```diff
@@ -0,0 +1,22 @@
...
+
+static EVENT_LOGGER: LazyLock<EventLogger> = LazyLock::new(EventLogger::new);
+
+wasm_msg_guest! {
+    fn track_event(request: Event) -> WasmResult<Void> {
+        EVENT_LOGGER.track(request);
+        Ok(VOID)
+    }
+
+    fn bounded_flush_events(_request: Void) -> WasmResult<PublishEventsRequest> {
```
nicklasl: I don't remember exactly how we handle the flushing of logs in the resolver but I think it makes sense to have the event flushing semantics intentionally match the resolver log flushing path? I think `bounded_flush_events` using the same bounded, drain-available approach as `bounded_flush_logs` makes sense, but I just want to make sure we’re not accidentally diverging on bounded vs unbounded flush behavior for the new WASM module.

openfeature-provider/proto/confidence/events/v1/events_api.proto
```diff
@@ -0,0 +1,31 @@
...
+option go_package = "github.com/spotify/confidence-resolver/openfeature-provider/go/confidence/internal/proto/events";
+option java_package = "com.spotify.confidence.sdk.events.v1";
+option java_multiple_files = true;
+option java_outer_classname = "EventsApiProto";
+
+message Void {}
+
+// A single tracked event, used both as WASM input and in the network batch.
+message Event {
+  // Event name, e.g. "eventDefinitions/my_event"
```
nicklasl: IMHO `eventDefinitions/` should be prepended by the wasm.

openfeature-provider/proto/confidence/events/v1/events_api.proto
```diff
@@ -0,0 +1,31 @@
...
+
+option go_package = "github.com/spotify/confidence-resolver/openfeature-provider/go/confidence/internal/proto/events";
+option java_package = "com.spotify.confidence.sdk.events.v1";
+option java_multiple_files = true;
+option java_outer_classname = "EventsApiProto";
+
+message Void {}
+
+// A single tracked event, used both as WASM input and in the network batch.
+message Event {
```
nicklasl: This looks like a Confidence events-shaped envelope, not an OpenFeature-shaped one?
OpenFeature track is basically:
>  - trackingEventName
>  - optional evaluationContext
>  - optional trackingEventDetails
>  - details include optional numeric value plus arbitrary custom fields

nicklasl: Maybe we have this as `PublishEvent` and have a separate OpenFeatureEvent?

openfeature-provider/proto/confidence/events/v1/events_api.proto
```diff
@@ -0,0 +1,31 @@
...
+  // Event name, e.g. "eventDefinitions/my_event"
+  string event_definition = 1;
+
+  // When the event occurred (set by the provider)
+  google.protobuf.Timestamp event_time = 2;
+
+  // Merged payload: user-supplied data with evaluation context nested under "context"
```
nicklasl: I don't think this applies to these providers.

wasm/event-guest/src/lib.rs
```diff
@@ -1,18 +1,66 @@
...
+fn build_payload(req: &TrackEventRequest) -> Option<Struct> {
+    let mut fields = std::collections::BTreeMap::new();
+
+    if let Some(data) = &req.data {
+        for (k, v) in &data.fields {
+            fields.insert(k.clone(), v.clone());
+        }
+    }
+
+    if req.value != 0.0 {
```
nicklasl: don't we need to be able to differentiate absence of a value with `0.0`?

wasm/event-guest/src/lib.rs
```diff
@@ -1,18 +1,66 @@
...
 use wasm_msg::WasmResult;
 
-const LOG_TARGET_BYTES: usize = 4 * 1024 * 1024; // 4 MB
+const LOG_TARGET_BYTES: usize = 2 * 1024 * 1024; // 2 MB
 const VOID: Void = Void {};
+const EVENT_DEF_PREFIX: &str = "eventDefinitions/";
 
 static EVENT_LOGGER: LazyLock<EventLogger> = LazyLock::new(EventLogger::new);
 
+fn build_payload(req: &TrackEventRequest) -> Option<Struct> {
```
nicklasl, line 22: Can we add a some tests and/or note + readme for the payload merge behavior here? I guess `value` and `context` intentionally win over same-named fields in `data`, but I just want to make sure that collision behavior is defined since OpenFeature custom data can contain arbitrary keys.

Dockerfile
```diff
@@ -129,6 +138,27 @@ FROM confidence-resolver.build AS confidence-resolver.test
 WORKDIR /workspace/confidence-resolver
 RUN make test
 
+# ==============================================================================
+# Build confidence-event-engine
+# ==============================================================================
+FROM rust-test-base AS confidence-event-engine.build
```
nicklasl, line 144: Can we also reference these new event-engine stages from the final `all` target further down in this file? I thing Docker won’t run this stage family in the main build unless something copies from `confidence-event-engine.test` / `.lint` there.

Dockerfile
```diff
@@ -219,6 +251,34 @@ FROM scratch AS wasm-rust-guest.artifact
 
 COPY --from=wasm-rust-guest.build /workspace/target/wasm32-unknown-unknown/wasm/rust_guest.wasm /confidence_resolver.wasm
 
+# ==============================================================================
+# Build wasm/event-guest WASM
+# ==============================================================================
+FROM wasm-deps AS wasm-event-guest.build
```
nicklasl, line 257: Can we also reference this event WASM stage from the final `all` target further down in this file? Docker won’t force `wasm-event-guest.build` / `.lint` / `.artifact` during the main build unless `all` copies from them.

openfeature-provider/go/confidence/internal/event_resolver/event_resolver.go
```diff
@@ -0,0 +1,207 @@
...
+	"github.com/spotify/confidence-resolver/openfeature-provider/go/confidence/internal/proto/events"
+	"github.com/spotify/confidence-resolver/openfeature-provider/go/confidence/internal/proto/wasm"
+	"github.com/tetratelabs/wazero"
+	"github.com/tetratelabs/wazero/api"
+	"google.golang.org/protobuf/proto"
+)
+
+// EventResolver wraps a WASM event engine instance and exposes TrackEvent
+// and FlushEvents operations using the wasm-msg protocol.
+type EventResolver struct {
```
nicklasl: Why is this called Resolver? `EventTracker`?

openfeature-provider/go/confidence/internal/event_resolver/event_resolver.go
```diff
@@ -0,0 +1,207 @@
+// Package event_resolver provides a WASM-based event engine for tracking
+// and flushing Confidence events via the wasm-msg protocol.
+package event_resolver
```
nicklasl: event_tracking?

openfeature-provider/go/confidence/internal/event_resolver/event_resolver.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	if er.runtime != nil {
+		err2 = er.runtime.Close(ctx)
+	}
+	return errors.Join(err1, err2)
+}
+
+// call implements the wasm-msg protocol: marshal request into a Request envelope,
+// allocate WASM memory, write, call the export, read the Response envelope, free.
+func (er *EventResolver) call(fnName string, request proto.Message, response proto.Message) error {
```
nicklasl: For these classes (this class and the resolver) where we do wasm integration we now have this duplication. Does it make sense to break it out in to an abstract class? (maybe that's not a Go thing, but anyway, code reuse)

openfeature-provider/go/confidence/provider.go
```diff
@@ -412,18 +470,157 @@ func (p *LocalResolverProvider) GetPrometheusMetrics(config SnapshotConfig) stri
...
+		return fmt.Errorf("failed to marshal events payload: %w", err)
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, eventsPublishURL, bytes.NewReader(jsonBody))
+	if err != nil {
+		return fmt.Errorf("failed to create events request: %w", err)
+	}
+	req.Header.Set("Content-Type", "application/json")
+
+	resp, err := p.httpClient.Do(req)
```
nicklasl: why did we choose http json here instead of grpc (which we use for the apply logs)?

openfeature-provider/java/src/main/java/com/spotify/confidence/sdk/OpenFeatureLocalResolveProvider.java
```diff
@@ -554,6 +596,186 @@ private ProviderEvaluation<Value> getObjectEvaluationInternal(
...
+        default:
+          if (c < 0x20) {
+            sb.append(String.format("\\u%04x", (int) c));
+          } else {
+            sb.append(c);
+          }
+      }
+    }
+    return sb.toString();
+  }
```
nicklasl: 😿

openfeature-provider/go/confidence/provider.go
```diff
@@ -125,17 +165,36 @@ func NewLocalResolverProvider(
...
 		statePollInterval:         statePollInterval,
 		logPollInterval:           logPollInterval,
+		httpClient:                httpClient,
 		enableApplyDedup:          options.enableApplyDedup,
 		disableExposureCollection: options.disableExposureCollection,
 	}
+
+	// Initialize event resolver if WASM bytes are provided
+	if len(options.eventWasmBytes) > 0 {
+		eventResolver, err := er.NewEventResolver(options.eventWasmBytes, options.useWasmInterpreter)
```
nicklasl: In the `Resolver` we have these "RecoveringResolvers" that wrap the resolvers and support reconstructing/restarting the WASM if a problem occurs. 
I don't expect us to salvage the events in a broken wasm instance but I think it makes sense to be able to keep the provider functioning.

openfeature-provider/go/confidence/provider.go
```diff
@@ -51,18 +59,44 @@ func WithLogPollInterval(d time.Duration) Option {
...
 }
 
-// WithEnableApplyDedup enables experimental apply-event deduplication in the
-// WASM resolver: repeated identical assignments within a short TTL window are
-// logged once. Off by default; the API may change.
+// WithEventTracking enables event tracking by providing the event engine WASM
+// binary. The WASM module must export wasm_msg_guest_track_event and
+// wasm_msg_guest_bounded_flush_events. Events are flushed on the same interval
+// as log flushing and POSTed to the Confidence events API.
+func WithEventTracking(eventWasmBytes []byte) Option {
```
nicklasl, line 68: I assume this is going to be called from the `provider_builder.go` and with that I expected the wasm file to be committed (needed by go:embed).
When we set that up, let's also make sure we have a similar check for the checked in wasm as the one we have for the resolve-wasm.

openfeature-provider/go/confidence/internal/event_resolver/event_resolver.go
vahidlazio: i didn't want to touch the resolver, since we might switch to state to wasm etc, but if we want to use same code, i'd do that in a separate PR.

nicklasl: sounds completely fair. didn't think about the upcoming plans.

openfeature-provider/python/src/confidence/provider.py
```diff
@@ -814,6 +919,149 @@ def _flush_assigned(self) -> None:
         except Exception as e:
             logger.error("Failed to flush assigned logs: %s", e)
 
+    def track(
+        self,
+        event_name: str,
+        context: Optional[EvaluationContext] = None,
+        value: Optional[float] = None,
+        data: Optional[Dict[str, Any]] = None,
+    ) -> None:
```
nicklasl, line 991: this does not match the [python provider interface declaration](https://github.com/open-feature/python-sdk/blob/main/openfeature/provider/__init__.py#L120-L125)

openfeature-provider/python/src/confidence/event_tracker.py
nicklasl, line 1: naming across this file -> event_tracker

openfeature-provider/python/src/confidence/event_resolver.py
```diff
@@ -0,0 +1,194 @@
...
+from wasmtime import Config, Engine, Linker, Module, Store
+from wasmtime import Trap as WasmTrap
+
+from confidence.proto.confidence.wasm import messages_pb2
+from confidence.proto.confidence.events.wasm.v1 import wasm_api_pb2
+
+logger = logging.getLogger(__name__)
+
+# Exception types that indicate a WASM crash requiring reload
+WasmCrashError = (RuntimeError, WasmTrap)
```
nicklasl: Could we make this narrower than all RuntimeErrors? I guess _consume_response also raises RuntimeError for a clean guest error envelope, so this path would reload the event WASM and drop the buffered instance even though the WASM may still be healthy.

confidence-event-engine/README.md
```diff
@@ -0,0 +1,89 @@
...
+  without flooding logs.
+
+Consequences worth knowing before relying on this for billing-grade data:
+
+- A hard failure (bad credentials, wrong endpoint) silently drops every event
+  for as long as it persists — only the periodic warning surfaces it.
+- Events buffered when the process dies uncleanly are lost. Shutdown drains up
+  to 100 batches, but a `SIGKILL` skips that entirely.
+
+## Known provider differences
```
nicklasl, line 64: When we ship this we should include this part in the main (as well as Go) readme.

openfeature-provider/go/confidence/internal/event_tracking/event_tracker_test.go
```diff
@@ -0,0 +1,116 @@
...
+		t.Fatalf("a decode failure must not be classified fatal, got %v", err)
+	}
+	if tracker.instance != before {
+		t.Error("instance was reloaded on a non-fatal error, discarding buffered events")
+	}
+}
+
+// A missing export means the module is not what we expect: that is fatal and
+// must reload.
+func TestFatalErrorReloadsInstance(t *testing.T) {
```
nicklasl, line 95: can we add tests that asserts that we can call correct functions before and after the crash?

openfeature-provider/java/src/main/java/com/spotify/confidence/sdk/LocalProviderConfig.java
```diff
@@ -180,6 +197,29 @@ public Builder disableExposureCollection(boolean disableExposureCollection) {
...
+     * {@link #eventWasmBytes(byte[])}.
+     *
+     * @param eventWasmPath path to the {@code confidence_event_engine.wasm} file
+     * @throws IOException if the file cannot be read
+     */
+    public Builder eventWasmPath(Path eventWasmPath) throws IOException {
+      this.eventWasmBytes = Files.readAllBytes(eventWasmPath);
+      return this;
+    }
+
```
nicklasl: why do we need to expose this? Why can't we use the same approach as for resolver wasm?

openfeature-provider/java/src/main/java/com/spotify/confidence/sdk/WasmEventTracker.java
nicklasl, line 1: WasmEventTracker.java

openfeature-provider/js/src/EventWasmResolver.ts
```diff
@@ -0,0 +1,134 @@
...
+    if (typeof exports[fnName] !== 'function') {
+      throw new Error(`Expected Function export "${fnName}" found ${exports[fnName]}`);
+    }
+  }
+  if (!(exports.memory instanceof WebAssembly.Memory)) {
+    throw new Error(`Expected WebAssembly.Memory export "memory", found ${exports.memory}`);
+  }
+}
+
+export interface EventResolver {
```
nicklasl:
```suggestion
export interface EventTracker {
```
and with all the rest of the renames that this would entail :)

openfeature-provider/js/src/ConfidenceServerProviderLocal.ts
```diff
@@ -214,11 +246,89 @@ export class ConfidenceServerProviderLocal implements Provider {
...
+    if (!this.eventResolver) return;
+    const batch = this.eventResolver.flushEvents();
+    if (!batch.events || batch.events.length === 0) return;
+    await this.sendEvents(batch, signal);
+  }
+
+  private async sendEvents(batch: FlushEventsResponse, signal = this.main.signal): Promise<void> {
+    const request = PublishEventsRequest.create({
+      clientSecret: this.options.flagClientSecret,
+      events: batch.events ?? [],
```
nicklasl: `?? []` shouldn't be needed or? we do that check in flushEvents

openfeature-provider/python/src/confidence/proto/confidence/events/v1/api_pb2_grpc.py
```diff
@@ -0,0 +1,128 @@
+# Generated by the gRPC Python protocol compiler plugin. DO NOT EDIT!
+"""Client and server classes corresponding to protobuf-defined services."""
+import grpc
+import warnings
+
+from . import api_pb2 as confidence_dot_events_dot_v1_dot_api__pb2
+
+GRPC_GENERATED_VERSION = '1.78.0'
+GRPC_VERSION = grpc.__version__
```
nicklasl: I believe this is generated and pointing to a different version compared to the already checked in flag logging grpc: https://github.com/spotify/confidence-resolver/blob/main/openfeature-provider/python/src/confidence/proto/confidence/flags/resolver/v1/internal_api_pb2_grpc.py#L8-L9

AI claims that this will not work in runtime?

openfeature-provider/js/tsdown.config.ts
nicklasl, line 35: @andreas-karlsson do you know if we need a change here as well? How can we pass the event-wasm to the inlined? or is the fix to include the new wasm in `openfeature-provider/js/src/index.inlined.ts`?

openfeature-provider/python/src/confidence/provider.py
```diff
@@ -335,6 +452,21 @@ def shutdown(self) -> None:
             except Exception as e:
                 logger.error("Failed to flush final logs: %s", e)
 
+        # Drain pending events. A single flush is capped inside the WASM, so
+        # anything beyond that cap needs further flushes or it is dropped.
+        if self._event_tracker is not None:
+            try:
+                self._drain_events()
```
nicklasl, line 453: Could we apply one overall shutdown deadline to the event drain, similar to Go and JS? I guess `_drain_events()` can enqueue up to 100 publishes and shutdown then waits for all of them? So a 30-second RPC timeout shutdown could block for quite a while during an outage.

openfeature-provider/java/src/main/java/com/spotify/confidence/sdk/OpenFeatureLocalResolveProvider.java
```diff
@@ -569,6 +666,137 @@ private ProviderEvaluation<Value> getObjectEvaluationInternal(
...
+    } catch (RuntimeException e) {
+      log.warn("Failed to flush events", e);
+    }
+  }
+
+  /**
+   * Drains all remaining events from the WASM engine by calling flush in a loop until no events
+   * remain.
+   */
+  private void drainEvents() {
```
nicklasl, line 736: Could we apply one overall shutdown deadline to this drain, similar to Go and JS?  `SHUTDOWN_GRACE` does not currently bound the total shutdown time?

nicklasl (commented):
There are no tests at all on the java provider and very few on the js provider.

openfeature-provider/go/confidence/internal/event_tracking/event_tracker.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+	return errors.Join(instanceErr, runtimeErr)
+}
+
+// call implements the wasm-msg protocol: marshal request into a Request envelope,
+// allocate WASM memory, write, call the export, read the Response envelope, free.
+// On a WASM trap the instance is reloaded and the error returned; the next call
+// runs against the fresh instance.
+func (t *EventTracker) call(fnName string, request proto.Message, response proto.Message) error {
+	t.mu.Lock()
```
nicklasl, line 137: Do we know how long the tracker lock is held when flushing a batch near the 2 MB limit? I guess it covers the WASM call, protobuf encoding, copying into host memory, and host-side decoding but not the network request. It would be useful to benchmark concurrent `track()` latency before deciding whether we need the same rotating-instance approach as the resolver.

nicklasl: And I guess the same question is for java as well. More of a discussion on if we should do pooling or not.
And if we decide to do pooling, wouldn't we as well just ship this in the resolver wasm?

vahidlazio: i think maximum couple of ms .. I don't see why it should be more.

openfeature-provider/js/tsdown.config.ts
andreas-karlsson: Yes. index.inlined needs a separate import for the event wasm.

nicklasl (approved):

PR #527, opened 2026-08-20
feat: disable exposure collection on local OpenFeature providers

vahidlazio:
## Summary

Adds a provider-level **`disableExposureCollection`** option on the local OpenFeature providers (Go `DisableExposureCollection` / `WithDisableExposureCollection()`, Java/JS `disableExposureCollection`, Python `disable_exposure_collection`, Rust `with_disable_exposure_collection()`).

When it is set:

- Apply / assignment / exposure logging is skipped **completely**. WASM never enqueues `FlagAssigned` on the assign queue.
- No deferred-apply resolve token is minted.
- Resolve logs (`client_resolve_info`, `flag_resolve_info`) and telemetry still flush.
- Assign flush loops are not started.

**Use only for exceptional no-exposure modes.** For normal feature delivery and experiments, keep applies enabled. When exposure collection is disabled, Confidence does not receive assignment/exposure events for those evaluations — experiment results, exposure counts, assignment diagnostics, and downstream reporting that depend on exposures can be incomplete or unavailable. Resolve analytics and telemetry are still sent, so this is not a general logging or privacy-off switch.

### How it works

Config `disableExposureCollection` is forwarded on every `setResolverState` as proto field 5 (`disable_exposure_collection`).

- **WASM providers:** stored in a guest-side static at `set_resolver_state`, then copied onto `AccountResolver` via `with_disable_exposure_collection()` for each resolve / `apply_flags`.
- **Native Rust:** kept as an instance field on `ConfidenceProvider` and stamped onto `AccountResolver` the same way.

Core resolve uses:

```text
if !disable_exposure_collection {
  match apply {
    true  → log_assign
    false → deferred token   // per-eval _confidence_skip_apply
  }
}
```

`log_resolve` always runs. `apply_flags` early-returns when exposure collection is disabled.

This is **not** the same as the existing per-evaluation `_confidence_skip_apply` context key. That path still sets `apply=false` and emits a deferred resolve token.

## How to use

Disable exposure collection for **all** OpenFeature evaluations:

**Go**
```go
provider, err := confidence.NewProvider(ctx, confidence.ProviderConfig{
    ClientSecret: "your-client-secret",
    DisableExposureCollection: true,
})
```

**Java**
```java
LocalProviderConfig config = LocalProviderConfig.builder().disableExposureCollection(true).build();
OpenFeatureLocalResolveProvider provider =
    new OpenFeatureLocalResolveProvider(config, clientSecret);
```

**JavaScript / TypeScript**
```ts
const provider = createConfidenceServerProvider({
  flagClientSecret: 'your-client-secret',
  disableExposureCollection: true,
});
```

**Python**
```python
provider = ConfidenceProvider(
    client_secret="your-client-secret",
    disable_exposure_collection=True,
)
```

**Rust**
```rust
let options = ProviderOptions::new("your-client-secret").with_disable_exposure_collection();
```

To skip apply for a **single** evaluation, pass `_confidence_skip_apply: true` in the evaluation context instead (deferred token path; config `disableExposureCollection` is unchanged).

## Tests

- Core: `disable_exposure_collection` + `apply=true` / `apply=false` (no token, no assign log, resolve logs still written); `apply_flags` is a no-op.
- Go / Java / JS / Python / Rust: resolve with `disableExposureCollection` still sends resolve infos and zero assigns.
- Java: `applyFlags` under `disableExposureCollection` is a silent no-op; packed int32 at proto tag 5 is skipped.
- Python: crash-reload restores `disable_exposure_collection`.
- JS: provider still flushes resolve logs on close and never calls `flushAssigned`.

## Note

`make sync-wasm-go` failed locally (Docker Desktop org sign-in). WASM was rebuilt with the local toolchain and copied into the Go asset. Re-run `make sync-wasm-go` after Docker auth if CI complains about WASM byte mismatch.


nicklasl:
I made some readme updates but I still don't like the param name. `skip_apply` sounds like “skip optional garnish.” The actual feature is more “disable exposure collection and make Confidence experiments/reporting sad.”

I’d prefer provider-level public API names like:

* JS: disableExposureCollection: true
* Go: DisableExposureCollection: true
* Java: .disableExposureCollection(true)
* Python: disable_exposure_collection=True
* Rust: .with_disabled_exposure_collection()

I’d keep `_confidence_skip_apply` as-is for the per-evaluation internal-ish context key because we don't want a broader migration. For the new provider config, `disableExposureCollection` is much clearer and appropriately ominous. It also avoids exposing the resolver’s “apply” terminology to users.

My vote: rename the public provider config before merge.

.github/workflows/ci.yml
```diff
@@ -34,6 +34,26 @@ jobs:
...
+          cache-from: type=registry,ref=ghcr.io/${{ github.repository }}/cache:main
+
+      - name: Upload WASM artifact (PR)
+        if: github.event_name != 'push'
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
+        with:
+          name: confidence_resolver.wasm
+          path: ./wasm-out/confidence_resolver.wasm
+          if-no-files-found: error
+
```
nicklasl: this shouldn't be here, right?

nicklasl (approved):

nicklasl (approved):

PR #528, opened 2026-08-20
fix: keep resolver string operators strict

nicklasl:
## Summary
- add resolver-spec cases for numeric starts_with and ends_with context values
- avoid numeric-to-string coercion for starts_with and ends_with targeting

## Tests
- jq empty confidence-resolver/test-payloads/resolver-spec/state.json
- jq empty confidence-resolver/test-payloads/resolver-spec/tests.json
- cargo test -p confidence_resolver --lib


nicklasl:
Superseded by #529, which uses the origin branch instead of the fork branch.

PR #572, opened 2026-08-31
feat(cloudflare): add event tracking via /v1/events:publish

vahidlazio:
## Summary
- Adds `POST /v1/events:publish` endpoint to the Cloudflare Worker — accepts both `application/x-protobuf` and `application/json`, so the JS SDK can swap the URL from `events.confidence.dev` to the CF worker with no code changes
- Validates `client_secret` against the deployed secret (401 on mismatch)
- Events are queued to a Cloudflare Queue (`events-queue`, batched at 100 entries / 10s), and the queue consumer ships them as JSON to `events.confidence.dev/v1/events:publish`
- Deployer script auto-creates the events queue on first deploy; prefix-aware like the flag-logs queue
- Always enabled — no opt-in required
- Returns 503 when queue binding is missing, propagates queue/serialization errors to client (no silent drops)

## Compatibility
| SDK transport | Supported | Notes |
|--------------|-----------|-------|
| JS (HTTP + protobuf) | Yes | Drop-in URL swap |
| JS (HTTP + JSON) | Yes | Drop-in URL swap |
| Go/Java/Python (gRPC) | No | gRPC is a different transport; CF Workers can't serve it |

## Test plan
- [x] 14 unit tests: JSON parsing, protobuf decoding (with nested Struct/Timestamp), event aggregation, auth validation, publish request structure
- [x] Deployed to `test-events-resolver` on Spotify Experimentation Cloudflare account
- [x] Verified 401 for wrong client secret (both JSON and protobuf)
- [x] JSON load test: 2000 events, 0 errors, 200 events/s
- [x] Protobuf load test: 2000 events, 0 errors, 3333 events/s
- [x] Verified events delivered through queue to backend
- [ ] Clean up test worker after merge

### DIFF
--- a/confidence-cloudflare-resolver/CLAUDE.md
+++ b/confidence-cloudflare-resolver/CLAUDE.md
@@ -20,17 +20,27 @@ A Cloudflare Worker that serves the Confidence flag resolver at the edge. Compil
 |--------|------|-------------|
 | POST | `/v1/flags:resolve` | Resolve flags (JSON body, `apply` respected when encryption key is available) |
 | POST | `/v1/flags:apply` | Apply flags (JSON body) |
+| POST | `/v1/events:publish` | Publish events (JSON body, queued and batched before delivery to backend) |
 | GET | `/v1/state:etag` | Returns deployment state ETag and resolver version |
 | OPTIONS | `*` | CORS preflight |
 
 Note: The Worker router uses `*path` matching because `:name` conflicts with Cloudflare router parameter syntax.
 
-## Queue Consumer
+## Queue Consumers
 
-The `#[event(queue)]` handler `consume_flag_logs_queue` processes batched flag log messages:
+The single `#[event(queue)]` handler dispatches on queue name:
+
+**Flag logs queue** (`flag-logs-queue`):
 1. Deserializes each message from JSON to `WriteFlagLogsRequest`
 2. Aggregates the batch via `flag_logger::aggregate_batch`
 3. Ships aggregated logs to the Confidence API
+
+**Events queue** (`events-queue`):
+1. Deserializes each message from JSON (array of Event objects)
+2. Aggregates all events from the batch
+3. Ships as a `PublishEventsRequest` to `events.confidence.dev/v1/events:publish`
+
+Both queues use `max_batch_size=100` and `max_batch_timeout=10s`.
 
 ## Environment Variables
 
--- a/confidence-cloudflare-resolver/src/lib.rs
+++ b/confidence-cloudflare-resolver/src/lib.rs
@@ -553,8 +553,8 @@ pub async fn main(req: Request, env: Env, ctx: Context) -> Result<Response> {
                     }
                     "events:publish" => {
                         let body_bytes: Vec<u8> = req.bytes().await?;
-                        let events = match parse_events_from_request(&body_bytes) {
-                            Ok(e) => e,
+                        let parsed = match parse_events_from_request(&body_bytes) {
+                            Ok(p) => p,
                             Err(msg) => {
                                 return Response::error(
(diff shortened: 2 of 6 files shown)


confidence-cloudflare-resolver/src/lib.rs
```diff
@@ -540,6 +551,65 @@ pub async fn main(req: Request, env: Env, ctx: Context) -> Result<Response> {
...
+                                .with_cors_headers(&allowed_origin);
+                            }
+                        };
+
+                        if let Some(expected) = CONFIDENCE_CLIENT_SECRET.get() {
+                            if parsed.client_secret != *expected {
+                                return Response::error("Unauthorized", 401)?
+                                    .with_cors_headers(&allowed_origin);
+                            }
+                        }
```
nicklasl: I wish we could do this before parsing the full payload.

confidence-cloudflare-resolver/CLAUDE.md
```diff
@@ -20,18 +20,28 @@ A Cloudflare Worker that serves the Confidence flag resolver at the edge. Compil
 |--------|------|-------------|
 | POST | `/v1/flags:resolve` | Resolve flags (JSON body, `apply` respected when encryption key is available) |
 | POST | `/v1/flags:apply` | Apply flags (JSON body) |
+| POST | `/v1/events:publish` | Publish events (JSON body, queued and batched before delivery to backend) |
```
nicklasl, line 23: JSON or proto

confidence-cloudflare-resolver/src/lib.rs
```diff
@@ -540,6 +551,111 @@ pub async fn main(req: Request, env: Env, ctx: Context) -> Result<Response> {
...
+                                            )?
+                                            .with_cors_headers(&allowed_origin);
+                                        }
+                                    };
+                                if probed != *exp {
+                                    return Response::error("Unauthorized", 401)?
+                                        .with_cors_headers(&allowed_origin);
+                                }
+                            }
+                        }
```
nicklasl, line 614: nice!

nicklasl (approved):

confidence-cloudflare-resolver/src/lib.rs
vahidlazio: Good catch — done in 78eec5d6, though with a caveat worth naming.

The secret is a body field, so a request that stays wire-compatible with the existing SDKs can't be authorized without touching the body *at all*. Two changes get as close as I could:

**1. Header fast path.** If the caller sends `Authorization: ClientSecret <secret>` — the convention `/metrics` already uses — the request is authorized and the body is never read. Wrong header → 401 immediately.

**2. Probe before full decode.** Otherwise, a struct declaring only field 1:

```rust
struct ProtoClientSecretProbe {
    #[prost(string, tag = "1")]
    client_secret: String,
}
```

prost hits `tag=2` (events), doesn't recognize it, reads the length varint and jumps past — so no `Event`, `Struct`, `Timestamp`, or payload `HashMap` is ever constructed. `parse_events()` runs only after auth passes. Correct regardless of wire field order, since prost scans the whole message either way.

**The caveat:** this avoids *allocation*, not *reading*. It's still an O(body) scan, and requests using body-secret auth now scan twice — once for the probe, once for the events. So it turns "unauthorized caller makes us build thousands of objects" into "unauthorized caller makes us scan bytes," which is real but less dramatic than it might read.

If you'd rather not pay the double scan on the happy path, I'm happy to drop the probe and keep just the header fast path with a single full parse — smaller diff, and unauthorized callers pay for the decode. Your call.

Also removed a 1 MiB body limit I'd added in the same commit (reverted in 29f2bb60) — it would have 413'd batches that `events.confidence.dev` accepts, which defeats the point of the endpoint.

PR #575, opened 2026-09-01
feat: add apply dedup and flush delivery telemetry

vahidlazio:
## Summary
- Add `ApplyDedupTelemetry` (proto field 10) tracking dedup counters: applies_total, applies_deduped, applies_not_cached, sweeps, map_size, map_capacity
- Add `flush_succeeded`/`flush_failed` (proto fields 11-12) tracking WriteFlagLogs batch delivery outcomes across all providers
- Host SDKs (JS, Java, Go, Python) track delivery locally and merge into TelemetryData before next send — zero WASM overhead, no resolve contention
- Cloudflare resolver tracks delivery in queue consumer

## Apply Dedup Metrics
| Metric | Type | Description |
|--------|------|-------------|
| `applies_total` | delta counter | Total flag applies entering dedup filter |
| `applies_deduped` | delta counter | Duplicates filtered out |
| `applies_not_cached` | delta counter | Passed through but not cached (map full) |
| `sweeps` | delta counter | Sweep operations executed |
| `map_size` | gauge | Current dedup map entries |
| `map_capacity` | gauge | Max allowed entries |

## Flush Delivery Metrics
| Metric | Type | Description |
|--------|------|-------------|
| `flush_succeeded` | delta counter | Successful WriteFlagLogs batch deliveries |
| `flush_failed` | delta counter | Failed WriteFlagLogs batch deliveries |

## Compatibility
Proto fields are additive — old consumers ignore them. Follow-up PRs needed:
- **epx-flags-admin**: `convertApplyDedup` + `convertFlush` in VictoriaMetricsClient
- **epx-flags-resolver**: same (has its own VictoriaMetricsClient)
- **epx-insights**: GraphType entries + PromQL queries for dashboards

## Test plan
- [x] 447 Rust unit tests pass (7 new tests for dedup counters, proto delta, Prometheus rendering, accumulate_delta, flush counters)
- [x] Clippy clean on confidence_resolver + confidence-cloudflare-resolver
- [ ] JS provider build
- [ ] Java provider build
- [ ] Go provider build
- [ ] Python provider proto regeneration
- [ ] Integration test with dedup enabled


confidence-cloudflare-resolver/src/lib.rs
```diff
@@ -704,11 +722,6 @@ async fn consume_flag_logs(
 
         let req = flag_logger::aggregate_batch(logs);
```
nicklasl, line 729: AI claims that the `aggregate_batch` method:
> merges only latency, resolve rates, state age, memory, and resolver version. Consequently, apply_dedup, provider flush, provider events, and incidentally provider_init_rate disappear before update_kv_snapshot or downstream delivery sees them.

openfeature-provider/js/src/ConfidenceServerProviderLocal.ts
nicklasl: I think `sendFlagLogsToDestination()` will return true even if the last retry fails which will mean that `sendFlagLogs` returns without throwing and therefore we increment and count that as a successful flush on L526.

This sounds like a mis-reporting.

openfeature-provider/go/confidence/provider.go
```diff
@@ -632,9 +632,15 @@ func (p *LocalResolverProvider) flushAndPublishEvents(ctx context.Context) int {
...
 
+	eventCount := len(batch.Events)
 	if err := p.publishEvents(ctx, batch); err != nil {
 		p.eventPublishFailures.Add(1)
 		p.logger.Debug("Failed to publish events", "error", err)
+		if p.flagLogger != nil {
+			p.flagLogger.RecordEventBatch(eventCount, false)
+		}
+	} else if p.flagLogger != nil {
+		p.flagLogger.RecordEventBatch(eventCount, true)
```
nicklasl: are we sure all events were accepted at this point? don't we need to look at the response?

openfeature-provider/java/src/main/java/com/spotify/confidence/sdk/OpenFeatureLocalResolveProvider.java
```diff
@@ -784,6 +795,7 @@ private void sendEvents(FlushEventsResponse batch) {
       }
```
nicklasl: Here we report successful publish even though we potentially got errors back

openfeature-provider/js/src/ConfidenceServerProviderLocal.ts
```diff
@@ -315,16 +320,20 @@ export class ConfidenceServerProviderLocal implements Provider {
...
         logger.error(`Failed to send events: ${response.status} ${response.statusText}`);
         return;
       }
+      this.eventBatchesSucceeded++;
+      this.eventsPublished += batch.events?.length ?? 0;
       const { errors } = PublishEventsResponse.decode(new Uint8Array(await response.arrayBuffer()));
       for (const error of errors) {
         logger.error(
           `Failed to publish event at index ${error.index}: ${EventError_Reason[error.reason]} ${error.message}`,
         );
```
nicklasl: same with other providers: we claim in telemetry that all events were published but some could have been rejected.

nicklasl (commented):
I think it makes sense to have a counter for rejected events.

openfeature-provider/go/confidence/flag_logger.go
```diff
@@ -7,4 +7,5 @@ import (
 type FlagLogger interface {
 	Write(request *resolverv1.WriteFlagLogsRequest)
 	Shutdown()
+	RecordEventBatch(eventCount int, succeeded bool)
```
nicklasl: This is a public API change. Is there a way to keep it internal?

openfeature-provider/proto/confidence/flags/resolver/v1/internal_api.proto
```diff
@@ -51,12 +51,65 @@ message TelemetryData {
...
+
+  message BucketSpan {
+    int32 offset = 1;
+    repeated uint32 counts = 2;
+  }
+
+  message ApplyDedupTelemetry {
+    uint32 applies_total = 1;
+    uint32 applies_deduped = 2;
+    uint32 applies_not_cached = 3;
```
nicklasl: could we name this a bit differently?
We increment this when the apply buffer "overflow", right? so maybe include something about that in the name?

openfeature-provider/js/src/ConfidenceServerProviderLocal.ts
```diff
@@ -503,50 +530,100 @@ export class ConfidenceServerProviderLocal implements Provider {
...
+    await this.serializeDelivery(() => this.deliverFlagLogs(writeFlagLogRequest, true, signal));
+  }
+
   private async flushAssigned(): Promise<void> {
     const writeFlagLogRequest = this.resolver.flushAssigned();
-    if (writeFlagLogRequest.length > 0) {
-      await this.sendFlagLogs(writeFlagLogRequest);
+    if (writeFlagLogRequest.length === 0) return;
+    try {
+      await this.serializeDelivery(() => this.deliverFlagLogs(writeFlagLogRequest, false, this.main.signal));
```
nicklasl: Could we avoid putting the per-evaluation assign path onto this single delivery chain? Every invocation of `evaluate()` calls `flushAssigned()` without awaiting it, so I think deliveries arriving faster than one network round-trip will build an unbounded queue and leave periodic telemetry flushes, state updates, and shutdown waiting behind it.

nicklasl (approved):

PR #579, opened 2026-09-02
feat(go): add WithApplyTime to backdate exposure events

MattiasMTS:
Server-side integrations often resolve a flag while handling a message whose facts are timestamped at arrival, a few seconds before the resolve. The exposure is stamped at resolve time, so the first treated message ends up on the pre-exposure side of `first_exposure`. That biases exposure-windowed metrics and contaminates the CUPED covariate; in one of our experiments it produced a +44% arm gap on a pre-randomization quantity and flipped a cost metric.

The protocol already supports this: `AppliedFlag.apply_time` is caller-supplied and `apply_flags` skew-adjusts it. The Go provider just can't set it, since `evaluate` applies inside the resolve and drops the token.

This adds `WithApplyTime(ctx, t)`. Evaluations made with that context resolve with `apply=false` and record the exposure via `ApplyFlags` with `t` as the apply time and the token from that resolve. `_confidence_skip_apply` and `DisableExposureCollection` still win. I went with a context value rather than an evaluation-context key like `_confidence_skip_apply` since the apply time is request-scoped rather than a targeting attribute, and it keeps the value typed; happy to switch if you prefer consistency with the existing key.

Go only for now. Also happy to move this into the core instead if you'd rather every provider get it in one place.

### DIFF
--- a/openfeature-provider/go/README.md
+++ b/openfeature-provider/go/README.md
@@ -623,10 +623,20 @@ value, err := client.BooleanValue(ctx, "my-flag.enabled", false, evalCtx)
 
 The key is automatically stripped from the context before it reaches the resolver.
 
+To record the exposure at a different time than the resolve, wrap the context with `WithApplyTime`:
+
+```go
+ctx = confidence.WithApplyTime(ctx, messageArrivalTime)
+value, err := client.BooleanValue(ctx, "my-flag.enabled", false, evalCtx)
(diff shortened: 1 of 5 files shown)


MattiasMTS:
Closing for now to align on the design with the team directly first; will reopen or replace once we've agreed on the shape.

PR #589, opened 2026-09-09
feat!: require encryption key

nicklasl:
Local providers currently allow callers to omit the encryption key and fetch plaintext state. Require a valid 64-character hexadecimal AES-256 key in JavaScript, Java, Go, Python, and Rust, reject invalid configuration before initialization, and remove plaintext fallback and public keyless construction paths.

Each provider has its own breaking-change commit, including API migration, encrypted fixtures, examples, and documentation. Shared commits update the benchmark server and integration guide. Migration instructions point to the unique encryption key alongside the matching client credential in [Confidence Admin](https://app.confidence.spotify.com/admin/clients).

ETags now advance only after state is successfully decrypted and loaded. Build validation also required retaining WASM assets in Python source distributions and enabling Tokio's signal feature for the Rust benchmark.

Validation:

- Python: 121 unit tests, Ruff, mypy, source distribution and wheel builds passed.
- JavaScript: 200 unit tests, typecheck, changed-source formatting, and build with locked dependencies passed.
- Rust: 113 unit tests passed, one existing test ignored; examples/E2E compilation and production-library Clippy passed.
- Java: 106 unit tests, wrong-key coverage, formatting, and package build passed.
- Go: unit tests, vet, demo/benchmark builds passed. Go and JS encrypted benchmark smoke tests completed with zero evaluation errors.

**BREAKING CHANGE:**  Passing an encryption key is now required when constructing the provider. Supply exactly 64
hexadecimal characters. Missing, empty, or malformed keys fail before network activity. The provider only fetches encrypted state and never falls back to plaintext.

Open [Confidence Admin → Clients](https://app.confidence.spotify.com/admin/clients), select your client, and locate the credential used by the provider. Each credential has its own unique encryption key, available alongside it. Use the key paired with your configured client secret. Configure and verify encryption on your existing SDK before upgrading.


openfeature-provider/go/CLAUDE.md
```diff
@@ -17,7 +17,7 @@ Go OpenFeature provider using the Confidence resolver compiled to WASM, loaded v
 ## Key API
 
 - **`NewProvider(ctx, ProviderConfig)`** (`provider_builder.go`) — Main factory function. Creates gRPC connection, state fetcher, flag logger, and wires everything together.
-- **`NewProviderForTest(ctx, ProviderTestConfig)`** — Factory with injectable `StateProvider` and `FlagLogger` for testing.
+- **`newProviderForTest(ctx, ProviderTestConfig)`** — Factory with injectable `StateProvider` and `FlagLogger` for testing.
```
vahidlazio, line 20: this is not unexported right? other places as well

vahidlazio:
i'd like to see some tests for this cases:

* Wrong encryption key** → decryption fails → ETag should NOT advance
* Tampered ciphertext** → authentication fails → ETag should NOT advance
* Truncated encrypted response** → decryption fails → ETag should NOT advance

vahidlazio (approved):

openfeature-provider/go/CLAUDE.md
nicklasl: Are you asking whether these are now unexported? They are. We made both helpers package-private to avoid public constructors that bypass the required encryption key. 
I guess the tradeoff is that consumers now need NewProvider with transport hooks for testing.

nicklasl:
> i'd like to see some tests for this cases:
> 
> * Wrong encryption key** → decryption fails → ETag should NOT advance
> * Tampered ciphertext** → authentication fails → ETag should NOT advance
> * Truncated encrypted response** → decryption fails → ETag should NOT advance

My take is that the existing failure-path tests cover what you have in mind. Go, Java, JS, and Python already check all three cases, including keeping the previous ETag for subsequent requests. Rust covers tampered and truncated responses, but I guess we should also connect its wrong-key test to the ETag assertions.

PR #590, opened 2026-09-09
feat: add support for managed evaluation context

vahidlazio:
## Summary
- Adds managed_context to Client proto (iam.proto + provider proto)
- New managed_context module: unflatten dot-paths to nested Struct, deep merge with managed values winning
- Integrated in ResolverState::get_resolver() - all consumers (WASM, Cloudflare, native Rust) get the merge automatically
- All 481 existing tests pass

## Merge semantics
- Managed app.name preserves request sibling app.version
- Managed scalar app replaces the entire request app subtree
- Managed values always win exact-path collisions

## Test plan
- [ ] managed_context module tests - 7 cases: empty, collision, nested merge, scalar replacement, all types, sibling preservation, unflatten
- [ ] All 481 existing tests pass (cargo test -p confidence_resolver --lib)


confidence-resolver/test-payloads/resolver-spec/state.json
nicklasl, line 1: could you fix so it doesn't re-format the whole file? :hurtrealbad:

confidence-resolver/test-payloads/resolver-spec/tests.json
nicklasl, line 1: same here, too many changes.

vahidlazio: created a separate shared managed context and spec test json and we reverted test and specs

nicklasl: I don't like that approach. That makes us have twice as many files to keep in sync :(

vahidlazio: the other option is to change a lot in test and spec files which are already out of sync.

confidence-resolver/src/lib.rs
```diff
@@ -48,6 +48,7 @@ mod bounded_set;
 mod err;
 pub mod flag_logger;
 mod gzip;
+pub(crate) mod managed_context;
```
nicklasl: do we need the (crate)?

nicklasl (approved):

confidence-resolver/src/lib.rs
vahidlazio: can be private

nicklasl (approved):

PR #607, opened 2026-09-16
fix(go): recover from startup state failures

### LINKED ISSUE #603: Go local provider should recover after initial CDN state fetch failure
## Problem

The Go local provider performs one CDN state fetch in `Init()`. A failure returns an initialization error before the state polling goroutine is started. The HTTP client has a 30 second timeout, so startup can block for that duration and the provider cannot recover automatically when the CDN becomes available.

## Intended behavior

Match the Java local provider startup behavior:

- Initialization should complete without waiting indefinitely for CDN state.
- The provider should remain not ready while no valid state is available.
- Flag evaluations should return the caller supplied default with a provider-not-ready error.
- Start state polling even when the initial fetch fails.
- Retry every 1 second while no valid state has been loaded.
- After a successful fetch, initialize the resolver, transition to ready, and use the configured normal polling interval.
- Shutdown must cancel in-flight requests and retry goroutines.
- Once state has been loaded, continue serving the last good state during later CDN failures.

## Acceptance criteria

- A startup CDN outage does not permanently fail provider initialization.
- Recovery happens automatically without recreating the provider.
- Tests cover timeout, repeated retry, recovery, default-value evaluation while not ready, and clean shutdown.
- The existing 30 second HTTP request timeout remains configurable through the existing transport hooks or an explicit provider option.

nicklasl:
## Summary

- keep the local provider not ready when the initial CDN state fetch fails
- retry every second until recovery, then resume the configured polling interval
- return defaults while not ready and cancel in-flight fetches during shutdown

## Testing

- `go test ./... -skip ^(TestFlagResolve_With|TestFlagLogs_Should)`
- focused startup recovery tests with `-race`
- `go vet ./...`

The remaining SDK readiness-reporting limitation is documented in the PR comment and addressed separately in #613.

Closes #603


nicklasl:
Resolved: #613 is now included in this PR. OpenFeature Go SDK v1.18.0 and the bounded initialization flow report startup errors and eventual recovery consistently, with regression coverage for synchronous/asynchronous registration and default/named providers.

vahidlazio (approved):

PR #613, opened 2026-09-18
fix(go): bound initialization with OpenFeature SDK v1.18.0

nicklasl:
Follow-up to #607, stacked on its branch.

OpenFeature Go SDK v1.16.0 can lose recovery events after an initialization error. Upgrade to v1.18.0 and retry startup state loading within one total 30-second budget, returning early on success or a recoverable timeout on expiry. Background retries continue after timeout, serving defaults until valid state is installed and readiness is reported.

Preserves the two commits separately: SDK upgrade, then bounded startup recovery. Includes SDK-level event-ordering and shutdown regression tests for synchronous/asynchronous default and named providers.

Validation: non-live Go tests, focused lifecycle tests with the race detector, and go vet.


nicklasl:
The SDK upgrade and bounded initialization changes are now included in #607. GitHub automatically marked this stacked PR merged when its base branch received those commits; further review continues in #607.

PR #615, opened 2026-09-18
feat(cloudflare): configure flag log queue count

nicklasl:
Cloudflare limits each queue to 5,000 messages/sec. Exceeding this causes `send()` to fail with `Too Many Requests`, visible in Worker logs as `flag log queue send failed`.

Adds optional `FLAG_LOG_QUEUE_COUNT` configuration to distribute flag logs randomly across multiple queues, all handled by the same consumer Worker. Defaults to one queue and preserves the existing queue name and binding. Use `FORCE_DEPLOY=1` when changing the count to bypass the unchanged-state deployment check.

Validated with 8 Rust tests, 20 deployer tests, and a WASM build. A five-queue deployment handled 5,000 resolves at 500 requests/sec for 10 seconds with `apply=true`; all requests matched successfully, and all five queues showed consumption and successful acknowledgements with no remaining backlog. Queue analytics are sampled.


vahidlazio:
you can do these learnt from [this](https://github.com/spotify/confidence-resolver/pull/616) :
1. Send-failure failover — #615 picks one queue and drops the log if it fails. #616's send_to_any_queue tries all remaining shards before giving up. Worth adding to #615 since the whole point of sharding is resilience under rate-limit pressure, and that's exactly when individual sends fail.
2. ensure_queue() helper in the deployer — #615 keeps the original inline queue creation for shard 0 and adds a loop for shard 2+. #616 extracts ensure_queue() which deduplicates the creation logic for flag-log shards AND the events queue (3 copies of the same curl pattern → 1 function). Not critical but cleaner.

nicklasl:
closing this in favour of #616

